PR #41, opened 2014-04-06
provider: don't crash when fetching bad URLs

fjullien:
Signed-off-by: Franck Jullien <franck.jullien@gmail.com>

### DIFF
--- a/bin/fusesoc.in
+++ b/bin/fusesoc.in
@@ -4,6 +4,11 @@ import os
 import sys
 import signal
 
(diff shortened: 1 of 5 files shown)


fjullien:
Do I close this ?

fjullien:
Might be related: http://pastie.org/9961644

olofk:
Thanks Franck! Both for the patch and for the patience. Had to resolve some conflicts, but I think I got it. Hopefully it will take less than ten months for me to apply future patches :)

PR #57, opened 2014-07-05
add simple bash-completion script for fusesoc

skristiansson:
Only the most basic commands are handled, but at least
it fetches the available cores/systems and allows for
autocompletion amongst them.

Further improvements would be to recognize sub-options.
(for instance, the common --elf-load should be easy to
implement as a file completion)


olofk:
This was taken care of with https://github.com/olofk/fusesoc/pull/197

PR #71, opened 2015-02-12
core: add local files to fetched core

fjullien:
Core can now be a mixture of remote files _and_ local files.
If a file exist in the remote location it can be overwritten by
the local version.

Signed-off-by: Franck Jullien <franck.jullien@gmail.com>


olofk:
Thanks. I applied the first commit. Was the second one pushed by mistake?

fjullien:
Yes, the second commit was made after I did this pull request. I didn't know it would be shown here.

PR #74, opened 2015-03-10
Update core.py

llessa:
The patch files list is not being correctly generated from the default "patches" directory


olofk:
Whoops! Thanks for reporting. I totally missed that. Applied and pushed now

PR #76, opened 2015-09-15
fixed problem with HTTPError

rinderer:


olofk:
Thanks for the pull request. I've applied the first patch but rejected the second since other changes need to be made insetad

PR #78, opened 2015-10-13
Minor review of README

mgielda:
Nothing fancy so far, just some cosmetics here and there.


olofk:
Thanks @mgielda Applied and pushed!

PR #80, opened 2015-10-16
Small changes to documentation files

jeffesquivels:
Fixed some typos, added some periods for consistency, updated a couple of links to their latest location and added what the acronyms in the readme stand for (hopefully this could help other people find the project through search engines).


olofk:
Thanks! Not entirely convinced that we need to write out all the acronyms, but I'm not against it either. Good find with the broken links as well. Applied and pushed!

jeffesquivels:
Thanks to you for the work on the project and for accepting my contributions!

PR #81, opened 2015-10-18
fix tostr method in launcher

mgielda:


olofk:
Thanks! Applied and pushed

PR #85, opened 2015-12-27
Use setuptools for the installation

chiggs:
For distribution of Python packages, using setuptools is preferred.  This enables the package to be hosted on PyPi, meaning that users can install using `pip`.

Note I haven't created a PyPi entry or tested, however a local pip installation works from the repo:

```
pip install .
```


chiggs:
@olofk I've re-based this, though with updated instructions to indicate the pip mechanism.  This does assume you would want to put the project on PyPi.

olofk:
Thanks for this @chiggs . I'll amend the change in bin/fusesoc and commit if that's ok with you

olofk:
The changes have been merged. Thanks for working on this

PR #86, opened 2016-01-25
For upstream

fjullien:


fjullien:
I added a commit here: https://github.com/fjullien/fusesoc/commit/a051bb2fcd523b27fad85a87c8471e42f9eb17aa

olofk:
Thanks! All three patches commited and pushed

PR #89, opened 2016-03-24
Fix type in error message when backend not found

wallento:


olofk:
Thanks for reporting. I actually noticed the same thing a few days ago, but it seems I never pushed the patch. Pushed my own patch since it fixed another thing as well

PR #92, opened 2016-03-28
Allow naming of files

wallento:
Hi,

these patches are part of the preparation for remote server providing cores. It generally makes sense to allow for names (we may enforce a format later). For now I use names like `oh/common/debouncer` or `wallento/pmod_switchbox` locally for testing a librecores remote API.

Cheers,
Stefan

### DIFF
--- a/bin/fusesoc
+++ b/bin/fusesoc
@@ -163,7 +163,8 @@ def list_cores(args):
     maxlen = max(map(len,cores.keys()))
     print('Core'.ljust(maxlen) + '   Cache status')
     print("="*80)
-    for name, core in cores.items():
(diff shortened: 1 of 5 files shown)


olofk:
Thanks for the patches. Having a sorted list of cores is a huge improvement. I'm never able to find what I'm looking for in the cores list :)

I need to think a bit about adding name as a parameter to the .core file. That was actually in the .core file until 29a7ea3744045e0. There are some drawbacks and benefits to doing it this way, and I need to make sure it won't conflict with future versioning ideas. I think it should be ok, but give me a few days to think about it.

wallento:
Closing, after discussion and converging on a different format.

PR #93, opened 2016-03-28
Add Vivado support

wallento:


olofk:
Looks like you forgot to add vivado.py :)

wallento:
lol, yes I did :) I rebased the commit.

fusesoc/build/vivado.py
```diff
@@ -0,0 +1,170 @@
...
+
+""" Vivado Backend
+
+The Vivado backend executes Xilinx Vivado to build systems and program the FPGA.
+
+The backend defines the following section:
+
+    [vivado]
+    part = <part name> # Format <family><device><package>-<speedgrade>
+    hw_device = <device name> # Format <family><device>_0
```
olofk: Do we need both?

fusesoc/build/vivado.py
```diff
@@ -0,0 +1,170 @@
...
+
+A core (usually the system core) can add the following files:
+
+- Standard design sources
+
+- Constraints: Supply xdc files with file_type=xdc
+
+- IP: Supply the IP core xdi file with file_type=xdi and other files (like .prj)
```
olofk: s/xdi/xci

fusesoc/build/vivado.py
```diff
@@ -0,0 +1,170 @@
...
+      as file_type=data
```
olofk: I would prefer to use one of the existing IP-XACT file types here, like 'use'r or 'unknown'. FYI, file types for latest revision are listed here http://www.accellera.org/images/xmlschema/ipxact/1685-2014/fileType.xsd

fusesoc/build/vivado.py
```diff
@@ -0,0 +1,170 @@
...
+        for s in src:
+            if s.file_type == 'xci':
+                ip.append(s.name)
+            elif s.file_type == 'xdc':
+                constr.append(s.name)
+            elif s.file_type.startswith('verilogSource'):
+                verilog.append(s.name)
+            elif s.file_type.startswith('systemVerilogSource'):
+                verilog.append(s.name)
+            elif s.file_type.startswith('vhdlSource'):
```
olofk: At least VHDL 2008 files need to be special cased here. Could perhaps be done in a follow-up commit

fusesoc/build/vivado.py
```diff
@@ -0,0 +1,170 @@
...
+
+{src_files}
+
+set_property include_dirs [list {incdirs}] [current_fileset]
+
+{xdc_files}
+
+create_run -name synthesis -flow "Vivado Synthesis 2015" -strategy "Vivado Synthesis Defaults"
+create_run implementation -flow "Vivado Implementation 2015" -strategy "Vivado Implementation Defaults" -parent_run synthesis
+
```
olofk: Do we need 2015 hardcoded here?

olofk:
Very nice and tidy. Probably has more comments than the rest of the FuseSoC code base, which is good :)

A few fixes and thoughts, but mostly ready to apply after that

fusesoc/build/vivado.py
```diff
@@ -0,0 +1,170 @@
...
+        # Write the formatted string to the tcl file
+        tcl_file = open(os.path.join(self.work_root, self.system.name+".tcl"), 'w')
+        tcl_file.write(PROJECT_TCL_TEMPLATE.format(
+            design       = self.system.name,
+            part         = self.system.backend.part,
+            bitstream    = os.path.join(self.work_root, self.system.name+'.bit'),
+            incdirs      = ' '.join(incdirs),
+            ip_files     = '\n'.join(['read_ip '+s for s in ip]),
+            src_files    = '\n'.join(['read_verilog '+s for s in verilog]+
+                                     ['read_vhdl '+s for s in vhdl]),
```
olofk: read_vhdl must be able to set the library too. Might be required for sv files too. See https://github.com/olofk/fusesoc/commit/0a3e895fb079fbe04d85a12cde9a2f12c5fd1c83 for reference how this is done in the ISE backend

PR #94, opened 2016-04-13
Sort cores alphabetically in list_cores

wallento:


olofk:
Thanks. Applied and pushed. I thought I had already picked that one

PR #95, opened 2016-04-13
Output filename in core config warnings

wallento:
Currently the core name is part of the warning. Use the filename
instead.

This is part of preparing a hierarchical naming scheme for remote
cores.


olofk:
Thanks! Picked and pushed!

PR #99, opened 2016-04-16
Generate dependency graph

wallento:
Usage: fusesoc gengraph <core>

It generates a dot file <core>.dot to be further processed with
graphviz.

### DIFF
--- a/bin/fusesoc
+++ b/bin/fusesoc
@@ -253,6 +253,19 @@ def update(args):
             except subprocess.CalledProcessError:
                 pass
 
+def depgraph(args):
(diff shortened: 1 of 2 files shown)


olofk:
Cool. Thanks. I'll need to hold this a bit though, since it depends on your sanitized name functions, which aren't in FuseSoC yet

wallento:
Oh, for sure. I missed this

olofk:
I like the functionality, but I'm pretty sure this one won't apply any more after the new dependency handler was introduced. I can take another look if you find time to rebase it, but please close it otherwise

Issue #104, opened 2016-06-26
windows installation problems


nturley:
I ran into a few bumps trying out fusesoc, this is what I tried on a windows 10 machine with python 2.7.8
```
>pip install fusesoc
...
Successfully installed fusesoc-1.4
>fusesoc init
'fusesoc' is not recognized as an internal or external command,
operable program or batch file.
```
looks like setup.py put a file named fusesoc in C:/Python27/Scripts. Let's try running that.
```
>python C:\Python27\Scripts\fusesoc init
Traceback (most recent call last):
  File "C:\Python27\Scripts\fusesoc.py", line 25, in <module>
    from fusesoc.build import BackendFactory
  File "C:\Python27\Scripts\fusesoc.py", line 25, in <module>
    from fusesoc.build import BackendFactory
ImportError: No module named build
```
It's confused. I decided to try inside of a virtual environment.
```
>python venv/Scripts/fusesoc init
...
ImportError: No module named yaml
```
That got further. That one looks like a missing dependency
```
>pip install pyyaml
>python C:\Python27\Scripts\fusesoc init
...
INFO:  FuseSoC is ready to use!
```

UPDATE:

I think I figured out the issue: https://python-packaging-user-guide.readthedocs.io/en/latest/distributing/#scripts

"Although setup() supports a scripts keyword for pointing to pre-made scripts to install, the recommended approach to achieve cross-platform compatibility is to use console_scripts entry"

I'm working on a fix

nturley:
I submitted a PR for this issue

olofk:
Thanks. I've seen the PR. FYI, I love to have this working on windows, but I'm not sure anyone has tested it that much, so I expect you to hit some problems after it's installed. I'm happy to review and discuss any windows-related patches, but as I don't have easy access to a windows machine, I can't really test things myself that easily

nturley:
Cross platform testing is a pain. Usually that's what CI is for. Have you thought about using travis or appveyor? I'll spend some time looking into it and you can see if you like it.

olofk:
Oh, that would also be very welcome. I've been kicking myself for not doing unit testing from the beginning. Could have caught a lot of unnecessary errors that way. But just build/install testing with travis (or similar) would be great.

nturley:
I broke this up with #106, #107, and #108. So at this point this issue is scoped to windows not being able to execute the "fusesoc" file in C:/Python27/Scripts.

olofk:
[...]

PR #105, opened 2016-06-26
use console_scripts instead of scripts

nturley:
this change addresses #104. This page explains more about it.
https://setuptools.readthedocs.io/en/latest/setuptools.html#automatic-script-creation

Instead of copying business logic to Scripts, setuptools automatically generates a python file that calls into an entry point and places it in scripts. On windows machines, it generates an exe shim. It should be a little more cross-platform friendly, and a little more maintainable.

I also added pyyaml as a dependency in setup.py so pip automatically installs it.

To make it work, I renamed bin/fusesoc to fusesoc/main.py. This might break your development workflow. I recommend using a virtual environment and calling
`pip install -e .`
That will add your development version of fusesoc to your path.

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,7 +1,0 @@
-engines:
-  pep8:
-    enabled: true
-
-ratings:
-   paths:
-   - "**.py"
--- a/bin/fusesoc
+++ b/bin/fusesoc
@@ -287,7 +287,7 @@ def run(args):
     # Run the function
     args.func(args)
 
-def main():
+if __name__ == "__main__":
     logger.debug("Command line arguments: " + str(sys.argv))
 
     parser = argparse.ArgumentParser()
@@ -353,6 +353,3 @@ def main():
         run(parsed_args)
     else:
         parser.print_help()
-
-if __name__ == "__main__":
-    main()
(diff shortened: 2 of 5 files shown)


olofk:
Thank you! This is an excellent addition. I've mostly done autotools stuff before and haven't really gotten to grips with setuptools yet, so it seems like I have a thing or two to learn. I want take a closer look at this before pulling it in, so don't expect any action here for a few days

nturley:
I'm going to remove adding the pyyaml dependency and open a separate PR for it. That way you can resolve the dependency issue without worrying about most of this change.

olofk:
It looks like you completely reverted the first patch, leaving no changes if both are applied. I pulled the pyyaml dependency already, so it would be great if you could update this PR with the console_scripts stuff

nturley:
Yeah I got into a fight with git and git won. I'll try to clean this mess up.

nturley:
okay. I think I fixed it. Now I think it has pyyaml, console_scripts, and the --yes parameter.

olofk:
Thanks! I missed the --yes patch, so I had already added my own version of it. The console_scripts stuff is applied though, and seem to work just fine

Issue #108, opened 2016-06-29
setup.py pyyaml dependency


nturley:
setup.py is missing a dependency to pyyaml. If you do not have pyyaml installed then init will fail.
```
>fusesoc init
...
ImportError: No module named yaml
```

olofk:
Thanks. Fixed in 13a7cd825765b3296ddc8090760c88b0fc411bd5

PR #109, opened 2016-06-30
add pyyaml dependency

nturley:
addressing #108


olofk:
Thanks! Picked and pushed

Issue #106, opened 2016-06-29
Add CI for smoke testing


nturley:
discussed in #104, we should add continuous integration to fusesoc. There's different levels of testing that could be added, but this specific issue only tracks basic smoke testing

CI should install and execute init, list-cores, list-systems, and update
CI should get OS coverage on a linux system and a windows system.
CI should test multiple versions of python on each OS. Python 2.7 and 3.5 at minimum.

olofk:
Thanks for the patches. Tests are successfully running with both Travis and Appveyor now

PR #110, opened 2016-06-30
add travis and appveyor config files

nturley:
Addressing #106 

You're going to have to go to the travis and appveyor websites and give them access to the repo before these will do anything. After you enable them, you can add badges to your readme.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,14 @@
+language: python
+python:
+  - "2.7"
+  - "3.2"
+  - "3.3"
+  - "3.4"
+  - "3.5"
+install:
+  - "pip install ."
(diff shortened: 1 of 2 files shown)


olofk:
Thanks! I have created appveyor and travis accounts now and added FuseSoC to both. Your patch however seems to depend on a '-y' switch to 'fusesoc init', which doesn't exist. I like the idea though, so I think this should be added

nturley:
oh shoot. I forgot about that. I added that flag in my ci branch but I didn't add it to this one.

olofk:
Many thanks. I added the -y flag and pushed this commit. It even looks like it works :)

PR #114, opened 2016-07-16
Fix failure when no parameters set

wallento:
When running FuseSoC without parameters the backends (e.g., verilator)
failed because the members like vlogparam are missing. Hence, create
them at init.

### DIFF
--- a/fusesoc/edatool.py
+++ b/fusesoc/edatool.py
@@ -37,6 +37,11 @@ class EdaTool(object):
         self.env = os.environ.copy()
         self.env['BUILD_ROOT'] = os.path.abspath(self.build_root)
 
+        self.plusarg    = {}
+        self.vlogparam  = {}
+        self.generic    = {}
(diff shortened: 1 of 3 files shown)


wallento:
I just found this actually breaks the params then. Let me add another fix later..

wallento:
That's the problem: https://github.com/olofk/fusesoc/blob/master/fusesoc/simulator/simulator.py#L66

Why is that needed?

wallento:
This should fix it properly

olofk:
Thanks for the patch. I'm planning to drop support for the old plusargs section after the 1.5 release however, so most likely only d5c0d8a5da08c5ee7f5b8862a47acc0438ec00f6 will be needed after that.

olofk:
ok, I take that back. It's broken right now and applying these fixes makes it work again. Any plusargs-related cleanup can go in after 1.5. Thanks. Applied and pushed

PR #115, opened 2016-07-17
Verilator: Get sim files not synth files

wallento:


olofk:
This is actually by design. With the old [verilog] section, all synthesizable files were used for verilator, and that's generally how it's used.

Looking at it now however, I understand that it goes against how the other simulators do it, even though verilator itself is a bit different. The big problem with changing that now however is that it would probably break most cores with a verilator backend since it would pick up a lot of non-synthesizable code.

PR #117, opened 2016-07-18
Run make with multiple parallel jobs

imphil:
Verilator runs gcc, which can be parallelized easily and produces great
speedups in compile time.
This patch uses NUMBER_OF_CPUS * 2 parallel jobs (if possible), a
commonly used default value.

### DIFF
--- a/fusesoc/simulator/verilator.py
+++ b/fusesoc/simulator/verilator.py
@@ -1,4 +1,5 @@
 import logging
+import multiprocessing
 import os
 import shutil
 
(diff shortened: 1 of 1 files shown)


imphil:
CI failure is infra fail, not caused by the code.

~~~
Build started
git clone -q https://github.com/olofk/fusesoc.git C:\projects\fusesoc
git fetch -q origin +refs/pull/117/merge:
fatal: unable to access 'https://github.com/olofk/fusesoc.git/': Failed to connect to github.com port 443: Timed out
Command exited with code 128
~~~

olofk:
Seems to work fine. Just tried with a small design, but I got around 40% speedup. Applied and pushed. Thanks!

PR #119, opened 2016-09-17
OpTiMSoC merge: add environment variable, Vivado/XSIM, list core root and path search order

wallento:


fusesoc/simulator/xsim.py
```diff
@@ -9,101 +9,150 @@ class Xsim(Simulator):
...
+                ip.append(s.name)
+            elif s.file_type == 'xdc':
+                constr.append(s.name)
+            elif s.file_type.startswith('verilogSource'):
+                verilog.append(s.name)
+            elif s.file_type.startswith('systemVerilogSource'):
+                verilog.append(s.name)
+            elif s.file_type.startswith('vhdlSource'):
+                vhdl.append(s.name)
+            elif s.file_type in ("CPP", "C"):
```
olofk: I see three problems with this:
1. This means that you can add DPI source by either setting them in the DPI section or adding them with file type set to C/CPP files. Don't like to have multiple ways to do this. Given the choice, I would prefer to list them only in the filesets, since we have more control over scope and usage there.
2. I'm using the IP-XACT file types when possible. The correct types would be cSource and cppSource
3. I can't see a way to create multiple DPI libraries here. I know at least one design I'm working on packaging that uses multiple DPI libraries. Might be ok to only have one for a first implementation though. Need to think about that

fusesoc/simulator/xsim.py
```diff
@@ -9,101 +9,150 @@ class Xsim(Simulator):
...
+            design       = self.system.sanitized_name,
+            toplevel     = self.top_module,
+            incdirs      = ' '.join(self.incdirs),
+            parameters   = parameters,
+            part         = part,
+            ip           = ipconfig,
+            src_files    = '\n'.join(['read_verilog '+s for s in verilog]+
+                                     ['read_vhdl '+s for s in vhdl])))
+
+        if self._has_dpi():
```
olofk: Why not just "if self.dpi_srcs:" ?

wallento: I wanted it to be more descriptive

fusesoc/build/vivado.py
```diff
@@ -0,0 +1,190 @@
...
+        tcl_file.write(PROJECT_TCL_TEMPLATE.format(
+            design       = self.system.sanitized_name,
+            part         = self.system.system.backend.part,
+            bitstream    = os.path.join(self.work_root, self.system.sanitized_name+'.bit'),
+            incdirs      = ' '.join(incdirs),
+            ip           = ipconfig,
+            parameters   = parameters,
+            src_files    = '\n'.join(['read_verilog '+s for s in verilog]+
+                                     ['read_vhdl '+s for s in vhdl]),
+            xdc_files    = '\n'.join(['read_xdc '+s for s in constr])))
```
wallento: Vivado takes care of it itself

fusesoc/build/vivado.py
```diff
@@ -0,0 +1,190 @@
...
+
+{src_files}
+
+{parameters}
+
+set_property include_dirs [list {incdirs}] [current_fileset]
+
+{xdc_files}
+
+regexp -- {{Vivado v([0-9]{{4}})\.[0-9]}} [version] -> year
```
wallento: Should add a comment there. The "stragies" are named with the tool version and as we don't want a fixed version in it, we always use the newest. Hence, some tcl magic..

fusesoc/build/vivado.py
```diff
@@ -0,0 +1,190 @@
...
+
+create_run -name synthesis -flow "Vivado Synthesis $year" -strategy "Vivado Synthesis Defaults"
+create_run implementation -flow "Vivado Implementation $year" -strategy "Vivado Implementation Defaults" -parent_run synthesis
+
```
wallento: Yeah, there always is a backlog of strategies, but it is not clear how long it goes and this approach always takes the current one. At least until they rename it entirely..

fusesoc/section.py
```diff
@@ -364,14 +364,16 @@ class XsimSection(ToolSection):
     def __init__(self, items=None):
         super(XsimSection, self).__init__()
 
-        self._add_member('xsim_options', StringList, "Extra Xsim compile options")
+        self._add_member('top_module' , str, 'verilog top-level module')
```
wallento: IIRC top_module is the DUT, but let me double check.

fusesoc/simulator/xsim.py
```diff
@@ -9,101 +9,150 @@ class Xsim(Simulator):
...
+        for s in src_files:
+            if s.file_type == 'xci':
+                ip.append(s.name)
+            elif s.file_type == 'xdc':
+                constr.append(s.name)
+            elif s.file_type.startswith('verilogSource'):
+                verilog.append(s.name)
+            elif s.file_type.startswith('systemVerilogSource'):
+                verilog.append(s.name)
+            elif s.file_type.startswith('vhdlSource'):
```
wallento: Why? What is their type name?

fusesoc/simulator/xsim.py
wallento: You are right, I think this is obsolete after I added DPI. Multiple DPI libraries is not that tricky, I need to think about it later.

The problem with fileset is that DPI sources need more options usually, like include paths and libraries, hence the extra section.

fusesoc/simulator/xsim.py
olofk: The type name is vhdlSource-2008 (local FuseSoC extension as they forgot to add this to IP-XACT). VHDL 2008 files need to be added with read_vhdl -vhdl2008 (Check ug901 for reference). I also realized now that this doesn't set the library name (available in s.logical_name) either. See the modelsim or xsim backend for reference

fusesoc/simulator/xsim.py
olofk: We probably need a DPI section, but I want to move as much file-related stuff as possible to filesets. Include files can be handled by setting the is_include_file attribute, and I see that there is a swObjectLibrary type in IP-XACT which might be meant for libraries

fusesoc/simulator/xsim.py
olofk: Fair enough

fusesoc/simulator/xsim.py
wallento: Okay, found it. For the project-based workflow this is the choice:

    set_param project.enableVHDL2008 1

fusesoc/build/vivado.py
```diff
@@ -0,0 +1,190 @@
...
+        tcl_file = open(os.path.join(self.work_root, self.system.sanitized_name+".tcl"), 'w')
+
+        ipconfig = '\n'.join(['read_ip '+s for s in ip])+"\n"
+        ipconfig += "upgrade_ip [get_ips]\n"
+        ipconfig += "generate_target all [get_ips]\n"
+
+        parameters = ""
+        for key, value in self.vlogparam.items():
+            parameters += "set_property generic {{{key}={value}}} [current_fileset -simset]".format(key=key, value=value)
```
wallento: I somewhat remember it was necessary, but can double check

fusesoc/section.py
```diff
@@ -364,14 +364,16 @@ class XsimSection(ToolSection):
     def __init__(self, items=None):
         super(XsimSection, self).__init__()
 
-        self._add_member('xsim_options', StringList, "Extra Xsim compile options")
```
wallento: I will add a new parameter type: vlogdefine. Similar as vlogparam it can set a define..

PR #120, opened 2016-09-23
Vivado

wallento:


fusesoc/build/vivado.py
```diff
@@ -0,0 +1,202 @@
...
+                ip.append(s.name)
+            elif s.file_type == 'xdc':
+                constr.append(s.name)
+            elif s.file_type.startswith('verilogSource'):
+                verilog.append(s.name)
+            elif s.file_type.startswith('systemVerilogSource'):
+                verilog.append(s.name)
+            elif s.file_type.startswith('vhdlSource'):
+                vhdl.append(s.name)
+                if s.file_type == 'vhdlSource-2008':
```
olofk: I think you still need to load vhdl 2008 files with `read_vhdl -vhdl2008 $file`, and you also need to know which library to put each of the VHDL files in. My suggestion would be to generate the complete tcl command in this loop instead of deferring it to line 93. Then you can do something like `tcl_file.write('read_vhdl -vhdl2008 -library {} {}'.format(s.logical_name, s.name)` for vhdl 2008 files, `tcl_file.write('read_verilog -sv {}'.format(s.name)` for SystemVerilog files etc.

wallento: No, thats both not necessary with the project flow. Vivado has evolved from this.

wallento:
Manually rebased and picked.

PR #133, opened 2016-12-30
Doc updates

Fatsie:
Document git provider and ignore generated html files.


olofk:
Thanks! Extra happy that you made patches for the documentation. It must mean that someone actually reads it, and that I should keep working on improving it :)

Both patches applied!

Issue #134, opened 2017-01-04
How to use coregen provider ?


Fatsie:
I am trying to use coregen from a core file using the coregen provider. This is the .core file:

    CAPI=1
    [main]
    name = freecores.github.io:t80:Z80Rom
    description = ROM for Z80 debug
    simulators = isim
    
    [provider coregen]
    script_file = syn/xilinx/coregen/Z80Rom.xco
    project_file = syn/xilinx/coregen/Atlys.cgp

For the t80 core I have the following depend in the [main] section:

    [main]
    ...
    depend = freecores.github.io:t80:Z80Rom

When trying to simulate the top core with isim the files are not generated:

    $ fusesoc sim --build-only freecores.github.io:t80:T80
    INFO:  Preparing freecores.github.io:t80:Z80Rom:0
    INFO:  Preparing freecores.github.io:t80:T80:0
    
    Running: /home/eda/software/Xilinx/14.7/ISE_DS/ISE/bin/lin64/unwrapped/fuse TestBench -prj isim.prj -o fusesoc.elf 
    ISim P.20131013 (signature 0xfbc00daa)
    Number of CPUs detected in this system: 4
    Turning on mult-threading, number of parallel sub-compilation jobs: 8 
    Determining compilation order of HDL files
    Parsing VHDL file "../src/freecores.github.io_t80_T80_0/rtl/vhdl/T80_Pack.vhd" into library work
    Parsing VHDL file "../src/freecores.github.io_t80_T80_0/rtl/vhdl/GBse.vhd" into library work
    Parsing VHDL file "../src/freecores.github.io_t80_T80_0/rtl/vhdl/T8080se.vhd" into library work
    Parsing VHDL file "../src/freecores.github.io_t80_T80_0/rtl/vhdl/T80_ALU.vhd" into library work
    Parsing VHDL file "../src/freecores.github.io_t80_T80_0/rtl/vhdl/T80a.vhd" into library work
    Parsing VHDL file "../src/freecores.github.io_t80_T80_0/rtl/vhdl/T80_MCode.vhd" into library work
    Parsing VHDL file "../src/freecores.github.io_t80_T80_0/rtl/vhdl/T80_Reg.vhd" into library work
    Parsing VHDL file "../src/freecores.github.io_t80_T80_0/rtl/vhdl/T80sed.vhd" into library work
    Parsing VHDL file "../src/freecores.github.io_t80_T80_0/rtl/vhdl/T80se.vhd" into library work
[...]

PR #135, opened 2017-01-09
Coregen copy file fix

Fatsie:
Comes out of discussion in #134.
Also includes small formatting change.


olofk:
Thanks! Both patches are applied

PR #136, opened 2017-01-16
Ghdl standard fixes

Fatsie:
I'm using ghdl 0.33 and it can't mix 2008 files with other standard files.


olofk:
I have unfortunately limited experience with ghdl and no good code to test this properly, but the patches look fine and I trust your judgement, so they are both applied. Thanks!

PR #140, opened 2017-02-25
Support running fusesoc comamnds from a file

ahmedalsawi:
Usage: fusesoc exec-script run.fusesoc

it runs commands (update, list-cores, etc) defined in run.fusesoc. it accepts empty lines and python-style comments.

### DIFF
--- a/fusesoc/main.py
+++ b/fusesoc/main.py
@@ -255,7 +255,7 @@ def update(args):
                 pass
 
 def exec_script(args):
-    filename = args.script[0]
+    filename = args.script
     pr_info( "Running fusesoc script: " + filename)
     with open(filename) as f:
         lines = f.read().splitlines()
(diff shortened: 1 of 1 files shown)


olofk:
Hi. Thanks for the patch. Technically, I don't think there's anything wrong with the patch, but I'm wondering a bit about the use case. Wouldn't it be just as easy to create a script to run these commands instead? Could you try to convince me that I'm wrong? :)

ahmedalsawi:
Thanks for looking into this. I guess you are right. it's just as easy but i thought it would be nice to have something like "modelsim -do" or "vivado -source". This way, i can call fusesoc once with a single file instead of source'ing a wrapper script(assuming wrapper script only contains fusesoc commands). it's just more explicit this way.

PR #144, opened 2017-03-20
Verilator: Command line args

wallento:
Add command line arg forwarding for Verilator. A core (the testbench
usually) can set a parameter like this:

    [parameter vcd]
    datatype = bool
    paramtype = cmdlinearg
    scope = public

It is forwarded as `--vcd=1` to the simulation then.


wallento:
This contains unmerged fixes for the dependency handling

olofk:
Picked and applied

PR #146, opened 2017-03-23
Return 1 if fusesoc fails

imphil:
Currently, if a synthesis fails, fusesoc still returns 0. This prevents
other tools from acting based on the success of this step, which is bad.


olofk:
Ganz klar. Picked and pushed! Thanks

PR #147, opened 2017-03-29
Replace pr_info, pr_warn and pr_err with standard logging module.

benreynwar:
I went through and replaced the pr_info, pr_warn and pr_err with the standard logger.info, logger.warning and logger.error functions.

I added a ColoredFormatter into utils.py, and a setup_logging function that is called from main.py so that the logging to stdout for the command line utility remains unchanged.

Why: Having one way of logging things makes it easy for reusing pieces of fusesoc.  Using the logging module means we can easily control whether it goes to stdout, to a log file, or both.


olofk:
Excellent! This has been on my TODO as well for several years now :)
A few extra features regarding logging that I want to see
- Disable log writing by default
- Better handle light backgrounds. yellow on white is unreadable
- Controllable logging level (e.g. only err, warn+err, warn+err+debug) to stdout
Don't need to do those in this PR, just mentioning. Will review patch later

olofk:
Thanks @benreynwar I rebased against master, squashed the two patches and applied. Seeems to work fine. Picked and pushed

PR #148, opened 2017-03-30
Replaced some old style logging that slipped through conversion.

benreynwar:
Apparently some slipped through yesterday.  I must have had a typo when double checking they were all done.


olofk:
All good. Picked and pused. Thanks!

PR #156, opened 2017-05-17
Do not improperly escape string vlogparams

imphil:
Currently, setting a parameter like this in fusesoc:

fusesoc <other stuff> --HOST_IF=usb3

will result in a Vivado project file with the following line in it:

set_property generic {HOST_IF="usb3"} [get_filesets sources_1]

This is interpreted by Vivado as a parameter HOST_IF with the value
"usb3" including the quotation marks, as indicated in the synthesis log:

        Parameter HOST_IF bound to: "usb3" - type: string

What actually was intended is to set the value to usb3 (without
quotation marks).

This patch removes the always-on quotation of vlogparams and leaves it
up to the backends to do proper passing of strings to their users.

I've tested this in Vivado and Verilator, I'm not sure about the other
backends.


olofk:
Thanks for reporting. I see the problem, and it's great to get rid of that ugly hack in edatool. We do need to move this handling to at least icarus (see https://github.com/steveicarus/iverilog/issues/139)

Before accepting this patch, there are three things I would like to have the fixed first:
1. Implement the quoting in the icarus backend
2. Test the other backends (I can help out with that)
3. Make sure the regression tests pass

imphil:
I've updated to implement quoting in icarus and also updated the regression tests, however I was only able to perform manual testing in Vivado and Verilator with a real design.

olofk:
ok, on the simulator side I've tried this now with icarus, modelsim and isim. Haven't got xsim and rivierapro available unfortunately. If you could try it with xsim, I'm ready to take a chance on rivierapro.

On the synthesis side quartus works fine, but I haven't had time to test the others, so ISE and icestorm needs to be tested

imphil:
How are you doing the testing? I should have at least ISE around to give it a try.

olofk:
Using this module

    module m #(parameter a=0) (input d, output q);
       initial $display("%s",a);
       assign q = d;
    endmodule

this core file

    CAPI=1
    [main]
    backend = ise

    [fileset s]
    files = p.v[file_type=verilogSource]

    [simulator]
    toplevel=m
    
    [parameter a]
    datatype=str
    paramtype=vlogparam
    
    [quartus]
    family = "Cyclone IV E"
    device = EP4CE22F17C6
    top_module=m
    
    [ise]
    family = spartan6
    device = xc6slx45
    package = csg324
    speed = -2
    top_module = m

For different sims, I run `fusesoc --cores-root=. sim --sim=<simulator> p --a=hello`
For different builds I run `fusesoc --cores-root=. build p --a=hello`, with different backend settings in the core file, and just check the synthesis logs. Not really an extensive test, but just making sure nothing strange happens to the string.

I tested ISE too now. Need to rebuild Icestorm to check that too

imphil:
xsim also looks good

imphil:
Btw, while we're at it: Looking at this piece of code https://github.com/imphil/fusesoc/blob/9b5d48e34242db620ff479b74536877367ec6537/fusesoc/simulator/icarus.py#L71: doesn't "false" also contain an "e"?

olofk:
Yes, but it evaluates to 0.0, so I left it that way. Looking at it now, I might have missed something. It could perhaps be clearer to explicitly set to 0/1 instead

olofk:
Icestorm needs quoting as well. Can you send an updated patch? I can make the changes here otherwise, but I will not have time right now

imphil:
Oh well, it seems this was more annoying than I hoped for :-) Icarus didn't have build tests where the actual parameters are passed, so I added those. In addition bools are now 0/1, and the conversion is centralized. Let's see how far this gets us.

olofk:
Thanks for working on this. Sorry for the extra annoyance. To make things a bit more complicated, I have completely rewritten the icarus backend in a local patch, but it's based on some other refactoring, so I can't push it now. Having extra tests will help making sure I don't mess up things during the rebasing. I'll review the patch tomorrow

olofk:
Thanks. Picked and pushed

PR #157, opened 2017-05-18
vivado: Make our run the default run

imphil:
When opening the fusesoc-generated project in the Vivado GUI it shows a
new and empty synth_1 and impl_1 run as default instead of the one
already finished.

This set uses the runs created by default in Vivado's create_project,
avoiding the manual creation of the runs. (I've left the code to do so
in the TCL file as a starting point for manual adjustment should
multiple runs be required.)

Before screenshot:
image

### DIFF
--- a/fusesoc/build/vivado.py
+++ b/fusesoc/build/vivado.py
@@ -169,14 +169,16 @@ set_property include_dirs [list {incdirs}] [get_filesets sources_1]
 
 {extras}
 
-regexp -- {{Vivado v([0-9]{{4}})\.[0-9]}} [version] -> year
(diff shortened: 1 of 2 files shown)


olofk:
That looks a lot cleaner. Would just like a second opinion from @wallento to see if there was a good reason for the old way. Otherwise I think we can remove the commented code. Users are allowed to add any extra TCL files they want by declaring them as tclSource anyway

wallento:
Oh, yeah, I noted that too. Thanks, lgtm

olofk:
Tests passed. Picked & pushed. Thanks!

PR #162, opened 2017-07-31
section.py: check for repeated _add_member() call

frantony:
Also drop repeated _add_member('pre_run_scripts', ...) call.

Signed-off-by: Antony Pavlov <antonynpavlov@gmail.com>


olofk:
Hi Antony,

Thanks for the patch. Looks correct. I would like to have a test case for this as well (one that would fail without this patch). Would you like to write one? Otherwise I can do it myself this time

olofk:
Pushed this patch now. Thanks!

PR #163, opened 2017-08-09
Fix passing of generics/parameters to Vivado

imphil:
Using set_property multiple times in Vivado overrides the previously set
value. In the case of generics/parameters, this caused only the last
parameter to be set during synthesis, all other parameters are ignored.


olofk:
Yikes, you're right. Need to add more testing against the actual tools :/
Thanks for this. Picked and pushed!

PR #171, opened 2017-09-23
Implement add-paths command

awygle:
I found myself wanting an easy way to add a core root path without manually editing the configuration file(s), so I implemented an add-paths command in the same vein as the list-paths command.

I hope this is useful. It's maybe not the prettiest patch - if you want me to refactor it in some way please let me know.

### DIFF
--- a/fusesoc/config.py
+++ b/fusesoc/config.py
@@ -32,13 +32,13 @@ class Config(object):
         xdg_config_home = os.environ.get('XDG_CONFIG_HOME') or \
                           os.path.join(os.path.expanduser('~'), '.config')
         config = configparser.SafeConfigParser()
-        config_files = ['/etc/fusesoc/fusesoc.conf',
+        candidate_files = ['/etc/fusesoc/fusesoc.conf',
                         os.path.join(xdg_config_home, 'fusesoc','fusesoc.conf'),
                         'fusesoc.conf']
 
-        logger.debug('Looking for config files from ' + ':'.join(config_files))
-        files_read = config.read(config_files)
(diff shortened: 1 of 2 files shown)


olofk:
Thanks for the patch! I agree that library management is a useful addition. I need to think a bit more about the implementation though. There are some things I would like to change right away, like replacing the print commands with logger.info or logger.warning. As this is adding a new command, we need to be very conservative, because this command needs to be supported in future versions to retain backwards compatibility, so I want to make sure this is the right way forward.

awygle:
Absolutely. The thing I most dislike about the approach in this patch is the text menu to select the config file to modify. One alternative might be to choose the current active config file and log a warning or info indicating which file is modified.

awygle:
Moved add-paths to library subcommand per Gitter conversation. Also moved list-paths since it seemed to go with the other. Changed the print commands in library add to be logger messages. Changed library add to modify the config file of narrowest scope rather than presenting a menu.

Now the thing I'm least certain about is the fact that a new cores_root option is created in the config file of narrowest scope, even if that config file doesn't contain a cores_root already. This seemed sane but also weird enough to be worth a warning.

olofk:
Thanks! I think this is almost ready to go now. Just a few more things
1. As you ended up with not using any interactive prompt, there is no need to move the python2 fix for input from init to main (makes the diff smaller)
2. Please restore the old list-paths command. I know that this is currently used, and must remain for that reason
3. I would love to see some test cases for the new code

awygle:
Closing in favor of improving fusesoc.conf as discussed on Gitter

PR #172, opened 2017-09-29
Add Cygpath handling for misbehaving subprocesses

cr1901:
The Python interpreter has a few versions on Windows:
* The `mvsc` one provided by the official installer and conda
* The `gnu` ABI version provided by `msys2`
* `cygwin`, a POSIX wrapper
* A `midipix` version in the future, which is a superior POSIX wrapper.

Depending on what version/package you use, you can end up in an environment (and _not_ just due to the `PATH` variable) where Python invokes a tool that doesn't understand Windows paths. This patch adds an option to force using POSIX style PATHs if necessary using the `--cygpath` flag. The `--cygpath` flag is optional, and `Provider`s must manually provide support for the flag.

There are many cases where the `--cygpath` flag isn't required, and might not even be useful. For instance, a user who installed the `msvc` build of Python might have only binaries on the path which don't support POSIX-style paths at all. This patch is meant to be a compromise to support users who don't have (or don't want) to install a POSIX environment on Windows, and those (like myself) who basically only use `mingw`/`gnu` ABI binaries.

### DIFF
--- a/fusesoc/config.py
+++ b/fusesoc/config.py
@@ -15,6 +15,7 @@ class Config(object):
     _init_done = False
 
     archbits = 0
+    cygpath=False
     def __new__(cls, *args, **kwargs):
         if not cls._instance:
             cls._instance = super(Config, cls).__new__(cls, *args, **kwargs)
--- a/fusesoc/main.py
+++ b/fusesoc/main.py
@@ -297,6 +297,12 @@ def run(args):
     else:
         config.archbits = 64 if platform.architecture()[0] == '64bit' else 32
         logger.debug("Autodetected " + str(config.archbits) + "-bit mode")
+    if sys.platform == "win32":
+        config.cygpath = vars(args)['cygpath']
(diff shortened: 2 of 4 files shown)


olofk:
Thanks for the patch. I have no real opportunity try this out for myself, but it looks like a solid patch and should not break anything, so I'm happy to apply if it helps for your case. Just one question, is the opencores provider the only place where this is an issue, or should cygpath ideally be used in other places as well?

olofk:
I would prefer some way to avoid importing the Config singleton into opencores.py, but it's no big deal. Can figure out how to refactor that later

cr1901:
>Just one question, is the opencores provider the only place where this is an issue, or should cygpath ideally be used in other places as well?

AFAICT, only the opencores (read: the `svn` provided by `msys2`) `provider` needs the option right now; `git` can tolerate Windows PATHs. I've been adding support on an as-needed basis.

Issue #168, opened 2017-08-29
Make all backends support user files


olofk:
In order to export arbitrary files to the backends, `file_type=user` should be used for those files. Currently only ise, modelsim, quartus and rivierapro supports `user` files. This needs to be implemented for the other backends as well

olofk:
Looks like this is fixed now with #173

PR #173, opened 2017-09-30
Support user files for Vivado, XSim, Verilator, ISim, IceStorm, Icarus, GHDL

hatimak:
Hi! This should add support for `file_type=user` to backends Vivado, XSim, Verilator, ISim, IceStorm, Icarus, GHDL, solving issue #168.


olofk:
Thanks! That seems to work just fine. Picked and pushed

PR #174, opened 2017-10-01
xci support for xsim, vivado dat files support

bqwer:
adds support for xsim simulation for projects containing vivado's xci cores
adds dat files support for both xsim and vivado

dat files are user data files used for memory initialization and other purposes
in my experience vivado is quite picky about user data files and xsim is even more fragile -- it needs files to be copied into simulation root directory

tested under linux and widows on medium sized project and small sized project


bqwer:
it's not well documented (well, it's not documented at all to be fair), but to be able to use xci cores in fusesoc's vivado projects one should put every xci file in separate directory. In other case vivado will exit with error, refusing to put different ip products in the same directory.

bqwer:
Travis fails because 2 new lines and new section present in project tcl template (to support dat files)

olofk:
Thanks for the patch. There are quite a lot of changes and I haven't gone through them all yet, but I have some comments to begin with

I will apply https://github.com/olofk/fusesoc/pull/174/commits/5f0de49cdb01b88f37274e4e1b9144f6e58ecc00 right away as that is unrelated to the other stuff and I'm leaving some comments on other things in the commits

fusesoc/edatools/xsim.py
```diff
@@ -55,19 +133,39 @@ def build_main(self):
         #Build simulation model
         args = []
         args += [ self.toplevel]
+        args += [ 'glbl' ]
```
olofk: I don't think we should add glbl here, as that is only used for some Xilinx IP cores. Instead, we can set `toplevel = real_top_testbench glbl` in the .core file to indicate that we need two simulation toplevels

fusesoc/edatools/xsim.py
```diff
@@ -55,19 +133,39 @@ def build_main(self):
...
+                    for line in f:
+                        lib_name = line.split('=')[0]
+                        if lib_name not in libs:
+                            args += ['-L', lib_name]
+                            libs.append(lib_name)
+
+        args += ['-L', 'work']
+        args += ['-L', 'unisims_ver']
+        args += ['-L', 'unimacro_ver']
+        args += ['-L', 'secureip']
```
olofk: These should probably be added to xsim_options instead

fusesoc/edatools/xsim.py
```diff
@@ -2,13 +2,16 @@
 from .simulator import Simulator
 import logging
 from fusesoc.utils import Launcher
+import time
```
olofk: Why is time imported here?

olofk:
Looking at it from another point of view, I think there are several things that this pull request tries to fix and I would like to look at those separately.

1. You need the ability to copy arbitrary files to the work root. This is why the `dat` files are introduced. I wonder if we can reuse the `user` file type instead and (a) copy files to work root with a script or (b) add a flag to files to make FuseSoC copy/move them instead. (a) is possible to do right now. I'm using a pre_build script myself in a few cores to copy files to work_root. Problem with this approach is that scripts are limited to a platform. With CAPI1 there is no way to run different scripts on windows and linux for example. For that reason, I think (b) could be of potential interest instead as a stand-alone patch. I'm thinking that it could be possible to specify in the .core file sometthing like `files = rtl/file.v[copyto=work_root]` or something like that

2. There's a lot of code to generate HDL files from the XCI files. It might make sense to rewrite the xsim backend instead to use project mode. I _think_ that would make it easier to handle xci files by just specifying them and let xsim automatically generate the necessary hdl for us. @wallento submitted a patch for running xsim in project mode last year, but I never applied it. Might make sense to revisit that now

fusesoc/edatools/xsim.py
bqwer: I think it can confuse user. Not everyone is familiar with `glbl` and that it needs to be specified as second top for correct simulation. But I get the point of doing it another way. Then it has to be documented well.

fusesoc/edatools/xsim.py
bqwer: Same here. This are common libraries that is included in Vivado generated scripts (tcl) when exporting IP. I was not aware of this libraries until I had to study generated sources. For user there is only way to figure out what libraries have to be specified --- analysis of all generated IP products. Maybe we should parse this tcl files the same way we parse ini.

bqwer:
I agree on reuse of `user` file type. I've thought of if. But implementation becomes more complex and it has to be agreed on. Strange options (from user point of view) are presented and this complicates things. Can we copy all user files to work_root by default? May it be more versatile?

I've searched for patch from @wallento and did not find if. Can you point it to me?
I didn't know that project mode is possible in xsim. All documents that I've found presented the same way to doing simulation --- generate output products from IP in project mode, process them by `xvlog` and `xvhdl`, then compile with `xelab` and finally call `xsim`. Maybe there is another way --- i'm interested in it.

Most troubles comes from another issue --- different Xilinx IP uses different libraries, sometimes several libraries. It is impossible to tell which ones. The only way --- parse IP output products.

fusesoc/edatools/xsim.py
olofk: I think you're right. It probably doesn't hurt to always have this file when building with xsim or isim

fusesoc/edatools/xsim.py
olofk: It's not enough with those libraries however. We also got simprims and XilinxCoreLib which might be needed. And then there's the equivalent libs for VHDL. I'm not sure how to best deal with this, but until we come up with a robust way I think it's better to have them in the .core files instead of pushing them to the FuseSoC code base

olofk:
I decided to implement the copyto tag as I realized that it actually solves a few other problems as well, so I think it's good to have. In your case it can be used for putting data files directly under the work root, and you can also use it to copy xci files to different directoreies under the work root and use them from there, so that you avoid messing up files in your source repos.

I'm attaching the patch I mentioned. xsim project mode is part of the patch. It is likely very outdated, but might give you an idea of what I mean. I haven't used it myself, but I think it's very similar to the regular Vivado projects

[0001-Add-Xilinx-Vivado-backend-and-rework-XSIM.patch.txt](https://github.com/olofk/fusesoc/files/1367143/0001-Add-Xilinx-Vivado-backend-and-rework-XSIM.patch.txt)

I still think we need a proper xci parser though, so that we can use the generated HDL files in other simulators that don't handle xci natively. This is actually implemented already in the logicore provider. It isn't very obvious, has some shortcomings and is also not very documented, but you can find some info in my [blog post about FuseSoC 1.5](olofkindgren.blogspot.com/2016/08/fusesoc-15.html)

Eventually, I would like to have a stand-alone xci parser that can be part of the not-yet-existant generator framework, but we need to find a good solution in the near-time

olofk:
copyto is implemented now. You can find it on the `copyto` branch. Need some doc and tests before I push it to master

bqwer:
I found last minute error in `xsim.py`. Shame on me. Closing this pull request and will be working on it. Will test `user` files

olofk:
The copyto functionality was just pushed to master together with documentation in https://github.com/olofk/fusesoc/blob/master/doc/capi1.adoc

PR #177, opened 2017-10-09
Don't output "Warning" twice

imphil:
Currently, I get output like this:

WARNING: Unknown section 'dpi' in 'xxx.core'
WARNING: Warning: Unknown item "top_module" in section "xsim" in yyy.core

This commit removes the second "Warning".


olofk:
Thanks. Looks fine to me

PR #180, opened 2017-10-15
Add "quit" statement to isim.tcl to end simulation

ThomasHornschuh:
Adding a quit statement to the isim.tcl will leave the simulator when the test bench runs out of stimuli (e.g. with stopping the clock as soon a desired simulation result is reached).
(Sorry for the whitespace change in line 23..24)

### DIFF
--- a/fusesoc/edatools/isim.py
+++ b/fusesoc/edatools/isim.py
@@ -20,8 +20,8 @@ class Isim(Simulator):
         (src_files, self.incdirs) = self._get_fileset_files()
         for src_file in src_files:
             if src_file.file_type in ["verilogSource",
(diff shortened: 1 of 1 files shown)


olofk:
Ah, right. That is probably a VHDL issue. I just tried running `fusesoc sim --sim=isim i2c` and that exited automatically, but that probably is because it is a verilog testbench that uses an explicit `$finish`. The VHDL equivalent wasn't introduced until VHDL2008, so it makes sense to add this for VHDL users. I'm taking the liberty to undo the whitespace changes.

Thanks for the patch!

olofk:
Aha. Those whitespace fixes were correct, so I kept them. Picked and pushed

PR #181, opened 2017-10-22
Add Support for Xilinx bmm Files for ISE

ThomasHornschuh:
bmm (Block Memory Map) type files are used by Xilinx tools to update the Block RAM contents in a generated bitstream without re-synthesis. For details how this work, check the Xilinx documentation, specifically for the data2mem tool.

See the commit comment for further details. 
Maybe the patch is not "production" quality, see also my remarks on limitations/issues in the commit comment. 

Thomas

### DIFF
--- a/fusesoc/core.py
+++ b/fusesoc/core.py
@@ -386,7 +386,9 @@ class Core:
             elif _bname == 'ise':
                 _files += _append_files(_b.tcl_files, 'tclSource')
                 _files += _append_files(_b.ucf_files, 'UCF')
+                _files += _append_files(_b.bmm_files, 'BMM')
                 del(_b.ucf_files)
+                del(_b.bmm_files)
             elif _bname == 'quartus':
                 _files += _append_files(_b.qsys_files, 'QSYS')
                 _files += _append_files(_b.sdc_files, 'SDC')
--- a/fusesoc/edatools/ise.py
+++ b/fusesoc/edatools/ise.py
@@ -68,6 +68,8 @@ quit
                 tcl_file.write('xfile add {}\n'.format(f.name))
             elif f.file_type == 'UCF':
                 tcl_file.write('xfile add {}\n'.format(f.name))
+            elif f.file_type == 'BMM':
+                tcl_file.write('xfile add {}\n'.format(f.name))
             elif f.file_type.startswith('vhdlSource'):
                 if f.logical_name:
                     if not f.logical_name in _libraries:
--- a/fusesoc/edatools/isim.py
+++ b/fusesoc/edatools/isim.py
@@ -20,8 +20,8 @@ class Isim(Simulator):
         (src_files, self.incdirs) = self._get_fileset_files()
         for src_file in src_files:
             if src_file.file_type in ["verilogSource",
-		                      "verilogSource-95",
-		                      "verilogSource-2001"]:
+                              "verilogSource-95",
+                              "verilogSource-2001"]:
                 f1.write('verilog work ' + src_file.name + '\n')
             elif src_file.file_type.startswith("vhdlSource"):
                 f1.write('vhdl work ' + src_file.logical_name + " " + src_file.name + '\n')
@@ -43,6 +43,7 @@ class Isim(Simulator):
         f2 = open(os.path.join(self.work_root,tcl_file),'w')
         f2.write('wave log -r /\n')
         f2.write('run all\n')
+        f2.write('quit\n')
         f2.close()
 
     def build_main(self):
--- a/fusesoc/section.py
+++ b/fusesoc/section.py
@@ -30,6 +30,7 @@ Example: data/mem_init_file.bin[copyto=out/boot.bin]
         'QIP',
         'SDC',
         'UCF',
+        'BMM',
         'tclSource',
         'user',
         'verilogSource',
@@ -522,6 +523,7 @@ class IseSection(ToolSection):
         super(IseSection, self).__init__()
 
         self._add_member('ucf_files' , FileList, "UCF constraint files")
+        self._add_member('bmm_files' , FileList, "bmm constraint files")
         self._add_member('tcl_files' , FileList, "Extra TCL scripts")
         self._add_member('family'    , str, 'FPGA device family')
         self._add_member('device'    , str, 'FPGA device identifier')


ThomasHornschuh:
Sorry, the commit from my last pull request (isim quit) is added again to this PR. I think it happened because I based the PR on the ise_quotes branch (which is ok, because the PR uses the ise.py from this branch).  I also not sure how to build a test. Should I add a "test core" to the /home/thomas/fusesoc/tests/cores directory?  Maybe you can give me some hints...

olofk:
Oh... it wasn't my intention to close this pull request. It happened automatically when I deleted the isequotes branch. Don't worry. I'll just cherry-pick your patch instead once I have looked at it

ThomasHornschuh:
Git can be tricky ;-)

olofk:
ok, I have some comments. It works as intended, but I want to drop `bmm_files` from the ISE section and just have the possibility to add them to a fileset. The reason being that we don't have the same flexibility for files defined in the tool-specific sections (e.g. scope/usage/copyto etc). We might want to use them for modelsim and isim in addition to ISE. The only reason that the other tool sections have files in them is because that's how I did it originally, and now I need to keep backwards compatibility with older FuseSoC versions.

I have also kind of forgot how BMM files work. Do you manually specify which bmm file to use when you instantiate a block RAM, or is the path to the block RAM described in the BMM file? I have some memory of an older design where I was forced to put the BMM files in the directory where I launched the tool (work_root in FuseSoC terms). In that case it might make sense to always copy them there. The drawback is of course if you have multiple bmm files with the same name that might end up overwriting each other. Do you have any insights here if that would be a potential problem?

wallento:
BMM is the ISE way of describing how the BRAM address space maps to positions of BRAM in the logic. It can be used to initialize the BRAMs accordingly. The new version in Vivado are MMI files (and the program is `updatemem`), which are generic xml and allow for multiple address spaces for different processor cores etc. Just played around with it yesterday. Writing MMI files can be troublesome, but there are workarounds.

IMO the files (BMM and MMI) must be generated by the fusesoc flow and can then serve as input to the other tools, or do I miss something?

ThomasHornschuh:
Hi,
first of all sorry for my late reply. I’m absolutely fine with the supporting bmm only in the files section. I started with the tool section in analogy to .ucf files, but at the I found it better in the files section for the same reason as you. 
Regarding the bmm workflow:  Stefan already outined the purpose of BMM files. For this reason they are two variants of a bmm: The first  variant contains only the mapping of adress spaces to the logical names of primitives (for example https://github.com/bonfireprocessor/bonfire-soc/blob/master/soc.bmm). When passed to Xilinx synthesis it creates finally a „back annotated“ variant of it, which contains the actual location of the block RAMs, data2mem can only work with this.
Here the example: https://github.com/bonfireprocessor/bonfire-soc/blob/master/soc_bd.bmm
(Normally this version should not be part of a git repo, it is only there by accident...)

When creating a design manually with RTL you need to create the bmm manually, it is therefore input to FuseSOC. Some Xilinx tools like core generators may also create them.

With Vivado unfortunately things change completely, I have not get a similar process running.

wallento:
Ah, I remember. Sorry, forgot that its also an input. The flow with Vivado is apparently simpler, except its broken when you don't use microblaze. But an acceptable workaround exists: https://www.xilinx.com/support/answers/63041.html

ThomasHornschuh:
he „broken when you don’t use microblaze“ is exactly the point. I checked my notes in the topic, and answer record 63041 was already on my list. Nevertheless thanks for your help.
I will check again if this helps me. The example in the link uses a mem file. The nice thing about data2mem is that it can directly read an elf binary. If created with a matching linker script, data2mem can be directly added to a Makefile.
Comming back to FuseSoC my ISE workflow is currently as follows:
* FuseSoC creates the bitstream and an assorted bd.bmm file
* my Makefile compiles the firmware, creates a new bistream with data2mem and directly loads it into the FPGA (with a board specific programmer which is a variant of xc3sprog).

For this to be working it would be very helpful to have some kind of possibility in FuseSoC to specify a directory for the output artifacts of the build process. But maybe this is better done accomplished with a post-build script.

For Vivado things a generally very different. I have the impression that Xilinx does not support pure HDL language designs well, they focus more on things like IP integrator. So I have changed my strategy for Vivado and have started to package my Bonfire processor (which is actually a RV32IM implementation) into a Vivado IP core. The interfacing is very similar to Microblaze (with AXI4 busses) and indeed it works really well together with Xilinx IP. But it is of course away from a completely open source design...

Nevertheless I‘m thinking about using FuseSoC instead of Vivado IP Packager. Do you have any ideas in this direction?

wallento:
@ThomasHornschuh about your last question: It is often discussed but not planned to do so called "top-level generation" by FuseSoC. I suggest you have a look at Kactus2.

About IP Integrator (IPI) I suggest you take the easy path and use "Add Module". It automatically instantiates the HDL core in the IPI and detects the interfaces properly. I suppose you should use AXI because otherwise its kind of useless, but you are already doing it.

`data2mem` is deprecated in Vivado and was replaced with `updatemem`. It also loads elf files into block ram and uses the new MMI format, which is much better.

Here is what I did just a few days ago (cannot share unfortunately):
- Instantiate picorv32 as module (symbol "RTL")
- Connect it up to Block RAM (say instance name is `ram`) and other peripherals
- Generate bitstream
- Use `write_mmi <bram-name>`, here `write_mmi ram` which outputs `ram.mmi`
- Fortunately it also gives you the command line to use with `updatemem`, copy and fill in your data, done!

A bit off-topic apparently, but hopefully useful to discuss the whole initialization flow for FuseSoC, IMO.

ThomasHornschuh:
@wallento:
Indeed I’m actually using a RTL module in the block design, IP packager is not really nice to use when the design is still work in progress. I’m thinking of an IP core for some form of “release version”. 

The disadvantage now with the Vivado version of my project is now that it is globbered with all the Xilinx specific stuff. The Spartan /  ISE Version is now completely “clean”: Just the RTL and the .core files. Most parts also can be simulated with ghdl. Recently I created a Cloud9 workspace, added ghdl and fuseSoC to it and was able to run testbenches in it. It took me half an hour of work. Maybe not really incredible usefull, but it shows the power and potential of FuseSoC.

One question to your example: How did you create the Block RAM? On Bonfire it is inferred from the vhdl code, because the boot ROM is local to the CPU and not exposed at the AXI4 busses

wallento:
`write_mmi` works on a netlist level, you will have to find the generated name in the netlist view after implementation. I think there is some pretty straight forward naming mechanism, so that you can generalize it a bit. I used the BRAM generator and AXI BRAM controller, there its simpler, because the name is set in IPI.

I agree that HDL is often the cleaner version, but IPI has a charme of quick bringup, especially on zynq platforms.

olofk:
I just noticed your vivid discussion and I remember now how BMM files works in an ISE flow with those weird auto-created _bd files. So, for ISE, it's pretty straight forward. There is still an open question if we should always copy the bmm files (to avoid polluting the source repo with the _bd file). To begin with, I think we can keep them where they are, and let the --no-export users clean up the _bd files themselves. The other thing is if we want FuseSoC to do the data2mem operation. It could be handy, but I think we'll wait a bit and just concentrate on getting the basic stuff done. It can still be done manually with a post_build_script.

For Vivado, it would be good to have something similar for the mmi files. As I haven't worked at all with those, I'm happy to receive patches from someone else.

There's still one other thing I'm not sure I get. Are the BMM files ever used in simulation? I vaguely remember converting an elf to some other format (probably using data2mem) that was consumed by modelsim, but I don't remember if the bmm files were ever involved. So, what is the standard procedure for doing what bmm files+data2mem does in simulations?

olofk:
No it hasn't. Mainly because I hadn't checked the bug tracker for a while and forgot about it. My apologies for that. Looking at the issue now I think it's fine to add your patch for this, but removing the changes to the ise section and just keeping it as a new file type. I'm happy to update and push your patch if you want

olofk:
I took the liberty to update your patch, remove some things and rebase it to master. Please take a look at it [here](https://github.com/olofk/fusesoc/commit/211601603460b1332cc86edff7b26c53b8b48303) and if you're happy with it I'll push it to master

ThomasHornschuh:
Just checked it with my build workflow (thanks to docker such things are easy): Perfect.
BTW: Is there any plan for a new "release" from fusesoc soon?

olofk:
Alright. Pushed to master now.

Yes, I'm almost ready to do a new release. Just need to add experimental CAPI2 support and write some documentation. Unfortunately I got a lot on my plate right now, so it might take a while

PR #183, opened 2017-11-18
Reverse the -r run command arguments.

jackgassett:
The run_options have to come after the module name for this to work. 
Options like the following have to come after the module name to work:
run_options = --stop-time=100ms --wave=simulation.ghw --fst=simulation.fst --vcdgz=simulation.vcdgz --vcd=simulation.vcd

https://ghdl.readthedocs.io/en/latest/Invoking_GHDL.html


olofk:
Whoops! Thanks for that. Picked and pushed

PR #184, opened 2017-11-25
Remove global singletons from codebase

awygle:
This pull request refactors the Config and CoreManager classes so they are no longer singletons, and are instead passed to the functions and objects which need them. This is intended to make testing and future development easier.

I might have gone a little overboard here - if you wanted CoreManager to remain a singleton, then only the first two commits are needed.

### DIFF
--- a/fusesoc/config.py
+++ b/fusesoc/config.py
@@ -12,7 +12,6 @@ logger = logging.getLogger(__name__)
 
 class Config(object):
     archbits = 0
-    cygpath=False
 
     def __init__(self):
         #TODO: Add option to load custom config file
--- a/fusesoc/config.py
+++ b/fusesoc/config.py
@@ -11,20 +11,11 @@ import os
 logger = logging.getLogger(__name__)
 
 class Config(object):
-    _instance = None
-    _init_done = False
-
     archbits = 0
     cygpath=False
-    def __new__(cls, *args, **kwargs):
-        if not cls._instance:
-            cls._instance = super(Config, cls).__new__(cls, *args, **kwargs)
-        return cls._instance
 
     def __init__(self):
         #TODO: Add option to load custom config file
-        if self._init_done:
-            return
         self.build_root = None
         self.cache_root = None
         self.cores_root = []
@@ -74,4 +65,3 @@ class Config(object):
         logger.debug('cache_root='+self.cache_root)
         logger.debug('cores_root='+':'.join(self.cores_root))
         logger.debug('systems_root='+self.systems_root if self.systems_root else "Not defined")
-        self._init_done = True
--- a/fusesoc/core.py
+++ b/fusesoc/core.py
@@ -41,7 +41,7 @@ class FileSet(object):
         return s
 
 class Core:
-    def __init__(self, core_file):
+    def __init__(self, core_file, cache_root, build_root):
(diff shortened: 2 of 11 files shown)


olofk:
FuseSoC is all about excess and living life to it's fullest, so going overboard is not an issue. Thanks for the patches. I've glanced through some of them quickly and they seem fine, but I'll try to give a proper review this weekend

olofk:
Some comments. Overall it looks great and makes the codebase tidier, but there are some things I would like to see changed:

1. This is the big one. I'm afraid the first patch will break the cygpath setting in the opencores provider. Not sure how to deal with that to be honest. It's a bit awkward to send the config all the way down to the providers for just this one setting in one provider. Would like to have some input from @cr1901 if this should be done for all providers when `--cygpath` is enabled or if it's specific to svn. I'm inclined to move the cygpath handling into core.py

2. Make cd2c243e82039d2c2e66cbc78b581a97dda11893 a part of 9d7de917d86c2fe1bab5b84f7c554d8e4c7a4db0

3. Make 747e74f2d9e1d50cc34238052443eb67360e903c a part of 238c287cabe79e9f94b85b5b31de9610e8724460

cr1901:
>Would like to have some input from @cr1901 if this should be done for all providers when --cygpath is enabled or if it's specific to svn.

TLDR: Revert my changes until I think of a better solution.

Long story: The more I think about it, the more I'm inclined to feel my solution is a hack.

My philosophy on dealing with the mess on Windows is something like this (substitute `X` for `cygwin`, `msys2`, `native`):
* If using `X`, use the `X` `python` and call out to `X` binaries as much as possible. If calling out to an `X` binary is not possible, try to constrain abstraction leaks as much as possible. See [this issue](https://github.com/m-labs/migen/pull/88) for an example.

One big abstraction leak when mixing binaries from various environments in Windows is path handling. In my case, the `msys2` `svn` doesn't really exist; the `svn` that `msys2` provides is a `cygwin`-like binary that only understands POSIX paths. An `msys2` `svn` would understand both POSIX and Windows paths. So to contain the abstraction leak, I think we should only add [platform-specific logic](https://stackoverflow.com/questions/35956180/how-to-determine-the-python-is-mingw-or-official-build) to the OpenCores builder only to handle path conversions.

awygle:
I've made the changes we discussed for Windows support. @cr1901 , can you test this to make sure it works for your usecase? It did the right thing as I understand it for both of my msys Pythons, as well as my two Cygwin Pythons, but you probably have a more complete understanding of the issue.

olofk:
Looks great. I'll push it as soon as I hear the results from @cr1901's smoke test

cr1901:
@olofk @awygle Worked fine. No issues downloading from OpenCores with @awygle's fix, which tbh I prefer to mine.

```
DEBUG: Using cygpath translation
DEBUG: None
DEBUG:     svn co -q --no-auth-cache -r 3 --username orpsoc --password orpsoc http://opencores.org/ocsvn/jt51/jt51/trunk /home/William/.cache/fusesoc/jt51_0
```

Second line with `None` reflects the `cwd` argument to the launcher. I assume that's okay?

olofk:
Awesome. Thanks for patches and testing. Picked and pushed!

PR #185, opened 2017-12-30
Bug fix: Remove class variables from CoreManager class

awygle:
I messed up when de-singleton-ing the CoreManager class, leaving several variables as class variables instead of instance variables.

This patch corrects that oversight and also creates a "common_cm" CoreManager in test_common.py so that all the existing tests can easily get the "default" test CoreManager.

### DIFF
--- a/fusesoc/coremanager.py
+++ b/fusesoc/coremanager.py
@@ -120,13 +120,11 @@ class CoreDB(object):
 
 class CoreManager(object):
     _instance = None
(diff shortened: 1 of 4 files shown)


awygle:
Python 2 was broken when I got here, I swear!

olofk:
Sorry for the delay. Python 2 has been broken since last year. Finally got around to fix it.
Patched is picked and pushed. Thanks!

PR #186, opened 2017-12-31
Add 'library' concept to config files

awygle:
This PR adds support for "libraries" as a concept in FuseSoC config files. Libraries are based off of Gentoo's repos.conf. They live in sections of the form [library.name] and (currently) have three properties, two of which are optional.

sync-uri is the URI source of the library. Currently this is assumed to be a Git repo. Required.
location is where to store the "cached" library - essentially, the clone target for the Git repo. It defaults to `$XDG_DATA_HOME/name`.
auto-sync specifies whether the library should be automatically updated by `fusesoc update`. Defaults to `true`. If set to `false`, the library can still be updated by passing its name explicitly to `fusesoc update`.

### DIFF
--- a/fusesoc/config.py
+++ b/fusesoc/config.py
@@ -19,6 +19,8 @@ class Config(object):
         self.cache_root = None
         self.cores_root = []
         self.systems_root = None
+        self.library_root = None
+        self.libraries = {}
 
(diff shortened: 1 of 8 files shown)


olofk:
Thanks for this. All picked and pushed now!

PR #190, opened 2018-01-27
Modelsim changes

ahmedalsawi:
PR fixes two issues with Modelsim simulation. I tried de2 and de0_nano with an old Modelsim (10.1d)

- Use Launcher to run vpi build instead of fusesoc_main.tcl. vsim tcl seems to have a problem with gcc warnings having format specifier (0x%08X)

- Moved 'run -all' to fusesoc_run.tcl to make sure vsim loads simulation top first. otherwise, it passes '-do run -all' first and fails to load top

That said, i am using 6 years old Modelsim and i need to check if they are fixed with new releases But these changes should work with any release.

### DIFF
--- a/fusesoc/edatools/modelsim.py
+++ b/fusesoc/edatools/modelsim.py
@@ -159,11 +159,14 @@ class Modelsim(Simulator):
                  cwd      = self.work_root,
                  errormsg = "Failed to build simulation model. Log is available in '{}'".format(os.path.join(self.work_root, 'transcript'))).run()
         if self.vpi_modules:
+            f = open(os.path.join(self.work_root, 'vpi_build.log'),'w')
             args = ['all']
             Launcher('make', args,
                      cwd      = self.work_root,
-                     stdout = open(os.path.join(self.work_root, 'vpi_build.log'),'w'),
+                     stdout =  f,
+                     stderr =  f,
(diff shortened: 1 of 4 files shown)


olofk:
Thanks! Running make from within the TCL file was a bit of a hack, but I never expected modelsim to try and execute the GCC output. Stupid TCL!

I think there is one slight regression. If a user wants to rerun the build stage manually from the exported build directory, they are now required to run `make`, but I'm fine with that. It's not like it was a perfect solution before this :)

Picked and pushed!

ahmedalsawi:
Thank you. that's good point. Do you think wrapper scripts (something like build.sh and run.sh) is good idea? I mean for all edatools backends. it will be consistent way to build/run from build directory even if flow was changed.

olofk:
Yes, the plan is to have a consistent interface, but using makefiles instead of build scripts as that makes windows-compatibility easier.

The makefiles will have three targets
build: to build the simulation model or FPGA image depending on tool type
run: to run simulations. Not used for synthesis tools and such
run-gui: Run simulator tool in GUI mode, if there is one.

There should probably be a build-gui or similar to fire up the synthesis/P&R tool in GUI mode as well.

The makefile will also contain the default values of the runtime arguments, but these will be possible to overwrite in the FuseSoC run functions.

You can take a look at the newly rewritten xsim backend as that most closely resembles this vision

PR #191, opened 2018-01-27
Local library support

awygle:
Provide support for local libraries in the new library sections config methodology.

Any library specified to `library add` with a `sync-uri` which exists on the filesystem, and without an explicit `--location`, will be detected as a local library. A local library will be stored in the config file with `location` set and `sync-uri` unset. `auto-sync` is ignored for local libraries.

The mismatch of argument name and what will end up in the config file is unfortunate, but to fix it would open some irritating edge cases in the argument parsing. Given our long-term plans I decided not to go that route - let me know if that's not OK.

### DIFF
--- a/fusesoc/config.py
+++ b/fusesoc/config.py
@@ -96,8 +96,11 @@ class Config(object):
                 logger.warn(_s.format(str(e), library))
                 continue
 
-            # sync-uri is non-optional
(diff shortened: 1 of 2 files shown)


olofk:
Works fine for the intended usecase such as `fusesoc library add dummy ../Desktop` . Agree that we can improve later. Picked and pushed! Thanks

PR #192, opened 2018-01-28
Support for using 'sync-type' to determine library provider

awygle:
This patch adds a 'sync-type' field to the library sections of the config file which specifies how to acquire the library. Currently two sync-types are supported, 'git' and 'local', with 'git' assuming a Git repository and 'local' acting much like `pip install -e` - that is,  simply using the contents of the provided directory directly. Down the line we would add 'hg', 'svn', or other backends as needed.

The Provider classes/modules which are used to fetch cores are expanded to allow managing libraries as well.

This patch is intended for merging sometime after 1.8.

### DIFF
--- a/fusesoc/config.py
+++ b/fusesoc/config.py
@@ -123,7 +123,7 @@ class Config(object):
         logger.debug('library_root='+self.library_root)
 
     def add_library(self, name, library):
-        from fusesoc.utils import Launcher
+        from fusesoc.utils import Launcher, _import
         if not hasattr(self, '_path'):
             raise RuntimeError("No FuseSoC config file found - can't add library")
         section_name = 'library.' + name
@@ -162,12 +162,11 @@ class Config(object):
         self.libraries[name] = library
 
         try:
-            provider_module = importlib.import_module(
-                'fusesoc.provider.%s' % library['sync-type'])
+            provider = _import(library['sync-type'], 'provider')
         except ImportError as e:
             raise RuntimeError("Invalid sync-type '{}'".format(library['sync-type']))
 
-        provider_module.PROVIDER_CLASS.init_library(library)
+        provider.init_library(library)
 
         with open(self._path, 'w') as conf_file:
             config.write(conf_file)
--- a/fusesoc/config.py
+++ b/fusesoc/config.py
@@ -7,6 +7,7 @@ else:
     import configparser
 
 import os
+import importlib
 
 logger = logging.getLogger(__name__)
 
@@ -102,10 +103,17 @@ class Config(object):
                 # sync-uri is absent for local libraries
                 sync_uri = None
 
+            try:
+                sync_type = config.get(section, 'sync-type')
+            except configparser.NoOptionError:
+                # sync-uri is absent for local libraries
+                sync_type = None
+
             self.libraries[library] = {
                     'location': location,
                     'auto-sync': auto_sync,
-                    'sync-uri': sync_uri
+                    'sync-uri': sync_uri,
+                    'sync-type': sync_type
                 }
 
         logger.debug('build_root='+self.build_root)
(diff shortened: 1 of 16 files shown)


fusesoc/config.py
```diff
@@ -149,13 +162,14 @@ def add_library(self, name, library):
...
-            try:
-                Launcher('git', git_args).run()
-            except RuntimeError as e:
-                logger.error("`library add` failed: " + str(e))
+        try:
+            provider_module = importlib.import_module(
+                'fusesoc.provider.%s' % library['sync-type'])
+        except ImportError as e:
+            logger.error("Invalid sync-type '{}'".format(library['sync-type']))
+            exit(1)
```
olofk: exit is only allowed in main.py. Reraise as RuntimeError("Invalid sync-type '{}'".format(library['sync-type'])) and catch it in main instead

fusesoc/config.py
```diff
@@ -149,13 +162,14 @@ def add_library(self, name, library):
...
+
+        provider_module.PROVIDER_CLASS.init_library(library)
```
olofk: Had forgotten about this PROVIDER_CLASS thing, and I think it's a bit unnecessary. No big deal, but we could do something like the _import function in main.py

fusesoc/main.py
```diff
@@ -155,9 +156,19 @@ def add_library(cm, args):
...
+    else:
+        provider = 'git'
+
+    if provider == 'local' and not args.location:
+        logger.info("Interpreting sync-uri '{}' as location for local provider.".format(sync_uri))
         library['location'] = os.path.abspath(sync_uri)
+        library['sync-type'] = 'local'
+    elif provider == 'local' and os.path.abspath(args.location) != os.path.abspath(sync_uri):
+        logger.error("Location '{}' does not match sync-uri '{}' for local provider. Consider specifying only sync-uri.".format(args.location, sync_uri))
+        exit(1)
```
olofk: Wouldn't it make more sense to skip sync-uri and only use location for local libraries?

fusesoc/provider/git.py
```diff
@@ -1,11 +1,30 @@
...
+import os.path
 from fusesoc.provider.provider import Provider
 from fusesoc.utils import Launcher
 
 logger = logging.getLogger(__name__)
 
 class Git(Provider):
+    @staticmethod
+    def init_library(library):
+        if not os.path.isdir(library['location']):
```
olofk: Perhaps we should notify the user somehow that we didn't clone the repo. Raise RuntimeError perhaps?

fusesoc/main.py
awygle: If we do this, then we'll break #191 - `fusesoc library add dummy ../Desktop` won't work. This was the best compromise I found.

fusesoc/provider/git.py
awygle: Changed this to re-raise any Launcher errors as RuntimeErrors. I figure we can specify that init_library raises RuntimeErrors regardless of how the provider is implemented.

fusesoc/config.py
awygle: Done

fusesoc/config.py
awygle: Done as a second commit

fusesoc/main.py
olofk: I see. It's unfortunate that sync-uri means different things for local and remote libraries. Maybe we need to go back to the drawing table and rethink the CLI slightly, but until then we do it like this

fusesoc/config.py
olofk: Awesome. I was thinking of suggesting to move it to utils.py and clean up the other references to PROVIDER_CLASS, but you had already done exactly that :)

fusesoc/main.py
awygle: The issue is somewhat fundamental, I think. For all other provider types, a sync-uri of some sort is required. Local is always going to be special.

Perhaps we could do `fusesoc library add git http://github.com/fusesoc/fusesoc-cores` and `fusesoc library add local ../Desktop`, making the sync-type a required positional argument? This would allow each provider to take different arguments, which could be good or bad.

olofk:
All good. Picked and pushed. Thanks!

PR #193, opened 2018-02-03
Dockerise tests

ahmedalsawi:
This is a small Dockerfile for fuseson-img. it's based on ubuntu:latest. it defaults to running tests but It can be used for development as well. small cavity: any files created in a container, they will have root permission. 

couple of scenarios where this comes handy:
- version compatibility. I had a problem with dev machine that had an old git installed. it didn't support "git apply --unsafe-paths". If there is a minimum version for a dependency package (iverilog, verilator), maybe we can build and deploy those into the container.

- for system tests. need a portable way to install edatools like modelsim, Altera and xilinx stuff. This is for installation only. licensing for commercial software would be still a problem specially if you have node-locked license.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,15 @@
+## Build fusesoc-img
 # docker build  -t fusesoc-img  .
-# docker run  --rm -it  -v $PWD:/fusesoc fusesoc-img  -c "pip install -e .;py.test"
+## Run tests in container and bolt.
+# docker run  --rm -it  -v $PWD:/fusesoc fusesoc-img
+## Run container with bash terminal for manual debugging
+# docker run  --rm -it  -v $PWD:/fusesoc fusesoc-img  -c /bin/bash
 
 FROM ubuntu:latest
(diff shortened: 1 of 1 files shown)


olofk:
Thanks for this. Good point with git versions not supporting unsafe-paths. I should do a check for this so other users doesn't hit the same issue. Maybe fall back to using `patch` as a plan B.

Regarding the docker stuff, I don't have any experience myself, so I would like to ping @wallento and @oleg-nenashev since they have been doing some Docker stuff with FuseSoC and other EDA tools as part of the [LibreCores CI project](https://github.com/librecores/librecores-ci)

oleg-nenashev:
We have images like https://github.com/librecores/docker-images/blob/master/librecores-ci/Dockerfile, which also bundle some extra tools needed for FuseSoC so that you can run a real build on them.

ahmedalsawi:
Wow that's exactly what i had in mind for Dockerfile. Thanks @oleg-nenashev @olofk

Issue #187, opened 2018-01-13
RFC: Templating Engines


wallento:
Hi,

this is a request for comments for a feature I need and want to implement. The goal is to being able to generate files during the FuseSoC run. I describe the proposed extensions below:

## Fileset

I think it makes most sense to just add file types that each correspond to an engine

    [fileset template]
    files = systemgen.sv
    file_type = template_jinja2

It may make sense to have an extension for templates, but that would add more problems in my opinions.

## Engines

I propose to support a set of python template engines and add them on-demand. [Jinja2](http://jinja.pocoo.org) seems very popular and I will build my stuff with this one.

## Flow

The normal FuseSoC flow, but instead of copying files, we feed the input + variables into the template engine and copy the output.

## Template Variables: Input

The user can supply template variables as json (or whatever you guys fancy more). The format is roughly like:

    [
      "vendor0:library0:module0": {
         "var": "value"
      }
      "vendor1:library1:module1": {
         "var": "value"
      }
    ]

The user can supply one or multiple files. All are parsed into a common variable container.

## Template Variables: Output

The template engine is called with the variables applicable to the module. It may make sense to map the other variables and parameters too, but I think we should start minimal.

## Other fancy stuff

I can think about a lot of stuff, but lets keep it simple for now.

What do you think? Would that be a good starting point? It would save me from adding an extra tool to run before fusesoc and I think it covers many use cases. I can go on and implement a prototype of that for further testing and commenting.

wallento:
This also relates to the generators by @benreynwar: https://github.com/benreynwar/fusesoc_generators

Other than those, this template engines in this RFC are deeper integrated into the flow, but obviously restrict the degrees of freedom a bit more. In my opinion they are complimentary.

imphil:
I like this proposal. Minor comments:
- Jinja2 is great. As a starting point, I've been using it here: https://github.com/imphil/fusesoc/tree/spyglass
- The common file extension for Jinja2 templates is ".j2", giving you editor highlighting in many editors.
- You need a way to specify the file type of the resulting HDL file. We have now some tools which need different code to add a SystemVerilog file vs. a Verilog file, or even special commands to add a SV-2009 vs. a SV-2001 file. I see two options:
  - Use the same file_type as today, "systemVerilog-2009" and invoke the templating logic whenever you detect a ".j2" file extension.
  - Use a combined file_type, e.g. "jinja2+systemVerilog-2009"
- I'd prefer YAML over JSON.
[...]

PR #194, opened 2018-02-28
Make use of Jinja2

imphil:
Introduce the basic setup for Jinja2 and make use of it for the Vivado backend.


imphil:
As usual the more tricky (and ugly) parts are related to Python packaging. New templates must be added to MANIFEST.in, using a wildcard match in package_data in setup.py doesn't work (it could for distutils, but not for setuptools). I gave up figuring out the exact details, that whole topic already consumed way too much time in my life in other projects.

olofk:
I hate Python packaging so much. It's an unholy mess. You got setup.py, setup.txt?, requirements.txt, MANIFEST.in and probably a dozen or so more files. I will do a review when I can, but you should also know that I have found enough problems in FuseSoC 1.8, so I'm rushing to fix some nasty bugs and do an 1.8.1 first

imphil:
no worries, this was laying around on my side for quite some time as well. I love those days when waiting for synthesis runs gives you time to code on unrelated things :-)

olofk:
https://xkcd.com/303/

wallento:
May be useful for #187. Once jinja is in the code base... ;)

olofk:
Status is that I completely forgot about it after the 1.8.1 release. Instead I started (and almost completed) a refactoring of all the backends and their test cases, which clashes with your patches. The good news is that by the time I had reached the letter v for vivado, I had lowered my ambitions considerabely, so I think the clashes are mostly confined to the test cases.

Regarding the patch, I see the benefits of using jinja and I'm happy to introduce it as a dependency. Just some details:

1. I'm not convinced about the is_*_file additions to edatool.py since some backends need explicit handling of which vhdl version is in use etc, and isim needs to treat vlog2005 as sv files, but I guess they can stay.

2. I'm moving all backends towards a model where the configure stage creates a makefile and the build stage just runs `make`. As an extra bonus for the ise backend I introduced an intermediate target so it's easy to just create the project and then open it in the GUI to do the build. I want to do the same for Vivado.

None of these items are critical to fix. Just mentioning them so that you're aware of some future work, and if you want to fix it you are most welcome. Otherwise, just rebase against master and I'll pick your patches

olofk:
One more thing. Instead of setting `include_package_data = True` in setup.py I think you should make a struct that looks like `package_data = {'fusesoc' : ['templates/vivado/vivado-program.tcl.j2', 'templates/vivado/vivado-project.tcl.j2']}`. IIRC this makes them available in both source distributions and binary distributions, but I might be wrong here

imphil:
Pushed through 74e96a22850daf402525a7c3cfb92cf685e18042 and 5ed3a927b1461356e9a9c10798f102125f92fe80

PR #198, opened 2018-03-01
Do not write fusesoc.log file by default

### LINKED ISSUE #195: Do not write fusesoc.log by default
Fusesoc by default writes a file 'fusesoc.log' into the working directory. Even though this is common for many EDA tools, I suggest removing this functionality by default.

Why?
- Having two fusesoc runs in the same directory results in an "interleaved" fusesoc.log file, essentially a useless one.
- It's very uncommon behavior for any "non-EDA" tool.

Two ideas how to resolve this:
- Write the fusesoc.log file to the build directory (`build/DESIGNNAME/...`)
- Don't write a fusesoc.log file unless instructed so by e.g. a command line option (e.g. `--logfile=fusesoc.log`)

imphil:
With a new option --log-file users can specify a log file if they want
to. Only in this case a log is written, otherwise all output is only on
STDOUT.

Fixes #195

### DIFF
--- a/fusesoc/main.py
+++ b/fusesoc/main.py
@@ -360,10 +360,10 @@ def update(cm, args):
             except RuntimeError as e:
                 logger.error("Failed to update library: " + str(e) + ". Continuing...")
 
-def init_logging(verbose, monochrome):
+def init_logging(verbose, monochrome, log_file=None):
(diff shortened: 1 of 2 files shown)


olofk:
Woohoo!! That only took about six years to get fixed :) Picked and pushed. Thanks!

PR #200, opened 2018-03-03
Library bugfix and --global flag

awygle:
Fixes local library detection bug and implements --global flag to indicate which fusesoc.conf should be used.

A possible enhancement might be to move the --global flag up to the `library` subparser or even the main `fusesoc` parser rather than leaving it at the `library add` level.


olofk:
All good. Picked and pushed. Thanks!

PR #202, opened 2018-03-26
Vivado: Use manual source management mode

imphil:
Since Vivado 2017.1 Xilinx has included a source code scanner into
Vivado which parses the HDL sources in order to display the hierarchy of
files in the UI. They did that by adding another parser for Verilog (at
least), which understands a different subset of Verilog than the parser
used in synthesis. As soon as non-parsable constructs are found Vivado
aborts the parsing process already before starting the "real" parser in
the Synthesis step, resulting in modules which cannot be found (even
though they are there and found during synthesis).

See https://www.xilinx.com/support/answers/69846.html for the Xilinx
take on this wonderful new feature.

We simply disable this new mode, since it doesn't bring anything except
increased build times and (sometimes) broken builds.

### DIFF
--- a/fusesoc/edatools/vivado.py
+++ b/fusesoc/edatools/vivado.py
@@ -102,7 +102,8 @@ class Vivado(EdaTool):
             parameters += "set_property verilog_define \""+defines+"\" [get_filesets sources_1]\n"
 
         if self.toplevel:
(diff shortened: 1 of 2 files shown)


wallento:
Thanks @imphil, this bug was killing me..

olofk:
Cool! Picked and pushed. Thanks

PR #208, opened 2018-05-04
Vivado: simplify programming TCL script

imphil:
The previously used template was unnecessarily complex. We anyhow only
support the standard case of a single board being connected to the local
machine, and in this case, the default set of TCL commands works just
fine (and also gives an appropriate error if the board is not found, for
example).

### DIFF
--- a/fusesoc/templates/vivado/vivado-program.tcl.j2
+++ b/fusesoc/templates/vivado/vivado-program.tcl.j2
@@ -2,27 +2,7 @@
 
 open_hw
 connect_hw_server
-
-set board ""
-
-foreach {{ target }} [get_hw_targets] {{ '{{' }}
-    puts "$target"
-    current_hw_target $target
-    open_hw_target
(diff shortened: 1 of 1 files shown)


olofk:
Simplifications are the best fixes! Picked and pushed. Thanks!

wallento:
I am not entirely sure if it isn't over-simplified: What about multiple hardware devices in the chain? For example in the zynq devices the fabric is not actually the first in the chain IIRC. I will assign that to me to check.

imphil:
Vivado takes only devices of the specified type. I didn't check Zynq devices, but having a VCU108 board and a Nexys 4 DDR board connected at the same time results in the right one being used and programmed.

wallento:
I see, but let me double check anyways

imphil:
Ah, I think I have found a case where the current version fails: if you have two (or more) boards connected and want to program not the first one. The previously used code wasn't fully working in my case either, I'll try to see if I can modify it in a way to work for this case.

PR #210, opened 2018-05-04
Vivado: find a HW target with the device

imphil:
The previous simplification of the Vivado programming TCL script removed
under some circumstances the ability to program a device if two devices
are connected to the PC.

STR:
- Connect two board with different FPGAs to the PC.
- Try to program both ones.

Before:
- Only the first board in the chain could be programmed.

Now:
- Both boards can be programmed.

Tested with a VCU108 and a Nexys 4 DDR board and Vivado 2018.1.

### DIFF
--- a/fusesoc/templates/vivado/vivado-program.tcl.j2
+++ b/fusesoc/templates/vivado/vivado-program.tcl.j2
@@ -1,8 +1,33 @@
 # Auto-generated program tcl file
 
+set bit "{{ bitstream_name }}"
+set device "{{ hw_device }}"
+
+puts "BITSTREAM: $bit"
+puts "DEVICE: $device"
+
 open_hw
 connect_hw_server
-open_hw_target
-current_hw_device [get_hw_devices {{ hw_device }}]
-set_property PROGRAM.FILE { {{ bitstream_name }} } [get_hw_devices {{ hw_device }}]
-program_hw_devices [get_hw_devices {{ hw_device }}]
+
+# Open all hardware targets to find one that contains the 
+# device we're looking for.
+set board ""
+foreach { target } [get_hw_targets] {
+    current_hw_target $target
+    open_hw_target
+    if { [get_hw_devices] == $device } {
+        puts "Found hardware target with a ${device} device."
+        set board [current_hw_target]
+        break
+    }
+    close_hw_target
+}
+if { $board == "" } {
+    puts "Did not find board with a ${device} device."
+    exit 1
+}
+
+current_hw_device $device
+set_property PROGRAM.FILE $bit [current_hw_device]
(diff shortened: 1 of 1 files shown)


imphil:
This addresses the comments by @wallento in https://github.com/olofk/fusesoc/pull/208

olofk:
What about two identical boards? I can't see a way to get around that other than with a runtime switch, but then we probably also need to match against some serial number to know which board we are targeting. All that is theoretical and complicated enough so I say we don't think about that right now.

Also, can we really match `$device` against `[get_hw_devices]` ? I'm guessing here that `get_hw_devices` will return a list of all devices on the JTAG chain, in which case, the check would fail if there are also other devices there. Shouldn't there be a `foreach { hw_device } [get_hw_devices]` inner loop? If we do that, I think we can also do `if { [get_property PART hw_device] == $part }` to get rid of the hw_device argument and use the `part` options that we need anyway. Haven't got any way to test this myself unfortunately

wallento:
Regarding @olofk first comment: Its actually not that complicated, and I have a draft lying around somewhere as part of our CI stuff.

About the second comment I agree and that was what I meant with my comment. On the zedboard there are two devices showing up, the DAP of the ARM cores and the fabric's FPGA. I will have a look at it this weekend. It may take us back to where we started ;)

olofk:
> Regarding @olofk first comment: Its actually not that complicated, and I have a draft lying around somewhere as part of our CI stuff.

If there's an API to get the serial number (and I guess there is from your comment), it will probably easy to match that. I'm not sure how to deal with that on the FuseSoC side though. But let's ignore that for now and solve the first problem.

> About the second comment I agree and that was what I meant with my comment. On the zedboard there are two devices showing up, the DAP of the ARM cores and the fabric's FPGA. I will have a look at it this weekend. It may take us back to where we started ;)

Right, I guess that would always fail to match `$device` against `[get_hw_devices]` since the latter is a list. I took a look now at the original implementation in FuseSoC. It iterated over targets, but never over hw_devices.

I just remembered that I have a Zynq device here too. Maybe I can find it and do some testing myself as well

olofk:
I think I made a proper fix now with https://github.com/olofk/edalize/commit/3f087d27323765ba656790d8d0cef8c433c70f2f
It still doesn't handle multiple identical devices, but I'm confident this can be fixed by adding a serial backend argument that matches any device if empty. With the recent option to supply backend arguments on the command-line this can even be usable without hacking the .core files, which is nice. Please reopen in edalize if this doesn't fix the issue

olofk:
Forgot to say. I based my work on @imphil's patch, but it got changed so much and transferred to another repo so I didn't bother to do the rebase dance and just mentioned this in the commit message

Issue #158, opened 2017-06-13
Support lint tools


imphil:
Currently fusesoc is pretty much centered around building a synthesis or a simulation. I've recently come across the need to also support linting with different tools, such as Verilator (using --lint-only) or commercial tools like SpyGlass. All these tools have in common that they also need the file/include list as it's generated by fusesoc.

Possible design options (or a mix of some of them):
1. Add a new "lint" class of tools, similar to "build" and "sim"
2. Remove the special-casing of build and sim tools, and make fusesoc more of a "tool runner". sim and build can remain "categories" of tools, however.

Orthogonal to that, the question of a "build sequence" comes up. Should one call to fusesoc do one step in a build process (such as "run synthesis" or "build simulation"), or could it also be a sequence of steps (such as "run lint first, then synthesize and then flash to FPGA").

olofk:
Yes, I've also been thinking how to add support for linters. It has been on my wishlist for a long time.  Thanks for bringing it up

1. One problem arises around which files to lint. Each tool can have tool-specific files. A way to solve that could be to instead add a --lint flag to the backend, such as there already are for --setup or --build-only

2. Yes. They are sharing more and more code, and some tools are actually a bit hard to categorize. I would need a good reason to make this change however. Do you have  any example?

There was a patch (https://github.com/olofk/fusesoc/pull/140) some time ago to do something similar. My reason at the time not to accept it was that I thought this could just as well be handled by a shell script. That would also make it easier to have conditions, such as only run synthesis if lint succeeds. I think you can already call most of the steps in the build process separately, even if it's not very clear how. --setup only runs the configure stage, --build-only runs configure+build, --keep will only run the simulation without rebuilding. I guess this could need some cleanup :)

imphil:
WIP of a SpyGlass lint implementation is available in https://github.com/imphil/fusesoc/tree/spyglass (Also featuring Jinja templating; I'll clean that up and split it out for #166).

Status:
1. It's a build tool, so it requires a separate *.core file with `backend = spyglass`. We need to find a solution here, in line with what was discussed above. 
2. It only writes the project file, doesn't perform the lint run yet
3. It doesn't allow to select a lint goal yet, I need to extend the `[spyglass]` section for that.

What's really blocking this now is the question how to integrate lint tools in general. Personally, I'd go with 
* Give up the distinction between simulation and build. Just make it an action. Each action roughly represents one tool run, e.g. "build with vivado", or "lint with verilator".
[...]

PR #211, opened 2018-05-18
Add Spyglass backend

imphil:
This PR adds a Synopsys (formerly Atrenta) Spyglass backend to fusesoc. The general approach has been discussed in #158. As discussed there the backend is CAPI2-only.

Example use
~~~
git clone git@github.com:optimsoc/optimsoc.git
cd optimsoc
git checkout spyglass
export OPTIMSOC_SRC=$PWD

fusesoc --verbose --cores-root $OPTIMSOC_SRC run --target=synth --tool=spyglass optimsoc:examples:compute_tile_nexys4ddr_capi2
~~~

While working on this I've noticed a couple things I don't yet understand of CAPI2 or might also be limitations. I've made note of these things in https://github.com/optimsoc/optimsoc/blob/spyglass/examples/fpga/nexys4ddr/compute_tile/compute_tile_nexys4ddr_capi2.core. In this case, all dependent cores are described in CAPI1 files, only the top-level core file is written in CAPI2. 

1. I cannot use a new (and self-defined) "lint" target to run spyglass, but I must use the synth target, otherwise no RTL files of our dependencies (as defined in CAPI1 files) are found.
2. I had to "hack" into the CAPI1 parser to actually add the files to the synth backend if Spyglass is used. 

I'm not sure how much of this logic is intended, what is a bug and what is just a limitation. Maybe @olofk you can give some insight here?

### DIFF
--- a/edalize/spyglass.py
+++ b/edalize/spyglass.py
@@ -1,0 +1,129 @@
+import logging
+import os.path
+import platform
+import re
+
+from edalize.edatool import Edatool
+
+logger = logging.getLogger(__name__)
+
+""" Synopsys (formerly Atrenta) Spyglass Backend
+
+Spyglass performs static source code analysis on HDL code and checks for common
+coding errors or coding style violations.
+
+Example snippet of a CAPI2 description file:
+
+spyglass:
+  methodology: "GuideWare/latest/block/rtl_handoff"
+  goals:
+    - lint/lint_rtl
+  spyglass_options:
+    # prevent error SYNTH_5273 on generic RAM descriptions
+    - handlememory yes
+
+"""
+class Spyglass(Edatool):
+
+    tool_options = {
+        'members' : {
+            'methodology' : 'String'
+        },
+        'lists': {
+            'goals': 'String',
+            'spyglass_options': 'String',
+        },
+    }
+
+    argtypes = ['vlogdefine', 'vlogparam']
+
+    tool_options_defaults = {
+         'methodology': 'GuideWare/latest/block/rtl_handoff',
+         'goals': [ 'lint/lint_rtl' ],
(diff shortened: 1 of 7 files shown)


edalize/spyglass.py
```diff
@@ -0,0 +1,113 @@
...
+            'has_systemVerilog' : has_systemVerilog,
+            'sanitized_goals'   : [],
+        }
+
+        self.render_template('spyglass-project.prj.j2',
+                             self.name + '.prj',
+                             template_vars)
+
+        # Create a single TCL file for each goal
+        goals = ['Design_Read'] + self.tool_options['goals']
```
olofk: FuseSoC crashes if goals is left out. Add a check and print raise a RuntimeError or set a sensible default if it's required

olofk:
> I cannot use a new (and self-defined) "lint" target to run spyglass, but I must use the synth target, otherwise no RTL files of our dependencies (as defined in CAPI1 files) are found.

Are you sure about this? I added toplevel, goals and spyglass_options from the synth target to the lint target and it looks to me like the synth files from CAPI1 cores are picked up

imphil:
Thanks @olofk for having a look. I've uploaded v2 of the PR.
- The synth vs. lint target issue is indeed non-existent, I don't know what I did wrong when I tested previously.
- I've added default values.

The example CAPI2 file at https://github.com/optimsoc/optimsoc/blob/spyglass/examples/fpga/nexys4ddr/compute_tile/compute_tile_nexys4ddr_capi2.core has been updated to showcase these changes.

olofk:
Looks great. I realize that I should have asked for a test case as well, but we can take that as a separate PR. Picked and pushed

PR #215, opened 2018-05-22
Run tests with tox

### LINKED ISSUE #213: Use tox for tests in travis
Currently fusesoc runs tests in travis with "pip install .". This prevents catching bugs which happen due to wrong packaging, i.e. when installing from sdist. Using tox first creates packages and then installs and tests them, making the test results more reliable in this regard. This would have prevented e.g. 165fa40bde15aea8f53eeddc20b87fcdc37f258a

imphil:
Using tox adds benefits over simply using "pip install ." in that it
- makes it easier for developers to run test on multiple Python versions
  (especially Python 2 and Python 3)
- and catches bugs which related to packaging (e.g. failing to add
  templates files to the distribution).

Fixes #213

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,12 +4,8 @@ python:
   - "2.7"
   - "3.4"
   - "3.5"
+  - "3.6"
 install:
-  - "pip install ."
-env:
-  - MODEL_TECH=dummy_value
+  - pip install tox-travis
 script:
-  - "fusesoc init -y"
-  - "fusesoc list-cores"
-  - "fusesoc update"
-  - "py.test"
+  - tox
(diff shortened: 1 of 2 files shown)


imphil:
Let's see how far this gets us once travis has finished its run.

imphil:
Not switching appveyor to tox for now since the system-level tests need more love. Travis/Linux works, however.

https://github.com/olofk/fusesoc/pull/215/commits/58cbe654b195254da9c082bef574763df671533d#diff-b91f3d5bd63fcd17221b267e851608e8R6 is rather ugly and "not the way to go", but it does the job for now. What would be better is to have the system tests (the fusesoc init, list-cores, etc. lines) in a separate script, or integrate them into pytest. Either way, checking the output would be a good thing (as opposed to just running them).

This goes well beyond this PR, so let's leave it as-is for now.

imphil:
Let's go with #440 instead.

PR #218, opened 2018-07-09
Add quartus library support

### LINKED ISSUE #217: Quartus backend support for compiling into multiple libraries
Hello! I'm currently experimenting with FuseSoC to see if I can make use of it to build some of our projects. I've noticed that the Quartus backend does not presently look for a logical_name associated with a file/fileset, meaning that projects which place files into separate libraries and then explicitly reference them fail to compile.

I have a local branch that I'm using to experiment with this functionality, which I could put up as a Pull Request if that would be helpful?

GCHQDeveloper992:
This updates the test_quartus test and backend implementation to allow Quartus to compile files into libraries based on FuseSoC's logical_name concept. Fixes #217


olofk:
Picked, squashed and pushed! Thanks!

PR #221, opened 2018-07-29
Pass the std to ghdl at elaboration phase.

euripedesrocha:
- Ghdl fails at elaboration if the VHDL source file is analyzed with --std=08.

The top entity cannot be found without the std parameter. Tested with ghdl 0.36-dev but seems to be the an old behavior (https://github.com/ghdl/ghdl/issues/160)


olofk:
Thanks! I finally got this taken care of. It seems to be the correct solution. I took the liberty to squash your commits into a single patch so that the test suite always passes. Picked and pushed!

Issue #219, opened 2018-07-11
A Quartus Pro backend


GCHQDeveloper992:
I'm currently working on projects that use both Quartus Std and Quartus Pro. The Pro version of the tools have made a few changes around the tool executable names and how QSYS gets invoked - this makes it incompatible with the current FuseSoC backend.

You mentioned in issue #217 that you were hoping to move backends toward the Vivado style templates, so I was considering trying to put something like that together for Quartus Pro. Whilst thinking about it I started wondering about how similar a Std and a Pro backend would look - I think the differences would end up being minor, probably something the templating could handle in a similar way to Vivado's use of the "{% if has_vhdl2008 -%}". 

Would you be happy with such an approach that performed some rough tool version identification in the backend (perhaps calling out to 'quartus_sh --version', defaulting to standard?), followed by an appropriate use of templates? Alternatively, would two distinct backends (perhaps with some shared code in a base class?) be cleaner, to avoid any direct dependency on tools? Given the incompatibilities between Std and Pro (QSYS files are not compatible between versions I believe), perhaps this second approach could be less risky?

olofk:
Apologies for the late reply.

This was very interesting and you raise many good points. My first instinct is that It would likely be easiest to have a single backend with some runtime detection, but it will require some thinking about corner cases. Incompatible qsys files sounds like it could cause some headaches, but perhaps there's a way to scan the files to see which version they are intended for and warn the user for incompatibilities. I don't have access to Quartus Pro so I'm a bit limited in testing. If needed I can see if I can use the trial version

I have a question about the differences in tool invocation. I know that the ip-generate command that is used by the current quartus backend is superseded by qsys-generate. Could it be that they have finally removed ip-generate from Quartus now and all that would be required is to change the code to call qsys-generate instead (optionally with a fallback to ip-generate for really old Quartus versions)?

GCHQDeveloper992:
Not a problem, I've not had a chance to look into this much myself yet! I do have a local branch which implements a separate QuartusPro backend, but I agree it would be nice to merge Std and Pro into one (I might tidy it up a bit and push it for discussion purposes).

[...]

PR #223, opened 2018-08-15
Add a Quartus Pro/Std backend

GCHQDeveloper992:
This addresses issue #219 by adding support for both Quartus Standard and Quartus Pro within the same backend. As discussed in that issue, this also moves the backend to the Jinja2 templating framework. I've tried to leave the end result of a Quartus Std run looking much as it did before this change.

Git seems to have confused a couple of the file moves and tried to track changes a little strangely, but the end result appears correct.

I've added tests to support checking both Pro and Std runs, but I haven't yet properly checked a set of real compilation runs - I intend to do that later this week. I thought I'd put this up to get feedback first though.


GCHQDeveloper992:
Looks like I've made a mistake with the Makefiles. I had a couple of issues with this locally and it seemed to be related to when the templates did or didn't give me a newline at the end of files. I'll try and take a look.

olofk:
Many thanks for this. I just wanted to say that I have unfortunately not had the time to review this yet, but it is high on my list of priorities. Thank you for the patience

PR #228, opened 2018-09-04
Do not output colors when redirecting output

imphil:
Unless passed --monochrome, fusesoc always outputs colors and bold text.
When redirecting fusesoc output this messes up the logs.

This change sets --monochrome by default when the output of fusesoc is
redirected, leading to an intuitive default behavior, in line with many
other tools.


imphil:
Looks like a CI infrastructure failure, not a test failure.

PR #229, opened 2018-10-18
Adding a Yosys+nextpnr+Trellis backend for the ECP5

daveshah1:


edalize/trellis.py
```diff
@@ -0,0 +1,57 @@
...
+                elif f.file_type == 'user':
+                    pass
+            for key, value in self.vlogparam.items():
+                _s = "chparam -set {} {} $abstract\{}\n"
+                yosys_file.write(_s.format(key,
+                                           self._param_value_str(value, '"'),
+                                           self.toplevel))
+
+            yosys_file.write("verilog_defaults -pop\n")
+            yosys_file.write("synth_ecp5 -nomux")
```
olofk: Do we want -nomux by default, or is this something that the user should provide when required?

fusesoc/capi1/section.py
```diff
@@ -490,6 +490,21 @@ def __init__(self, items=None):
...
+
+    TAG = 'trellis'
+
+    def __init__(self, items=None):
+        super(TrellisSection, self).__init__()
+
+        self._add_member('nextpnr_options', StringList, "nextpnr options")
+        self._add_member('yosys_synth_options' , StringList, "Additional options for the synth_* commands in yosys")
+        self._add_member('top_module', str, 'RTL top-level module')
+        self.top_module = 'orpsoc_top'
```
olofk: We don't want orpsoc_top as the default top_module. This is only a remain in some of the backends for backwards compatibility

edalize/trellis.py
daveshah1: At the moment, and for the near future, nextpnr does not support the LUT extension muxes, so disabling them seems best for now

fusesoc/capi1/section.py
daveshah1: Ah, think I added this while trying to get the old SoC project to build.

olofk:
Picked and pushed. Thanks!

Issue #188, opened 2018-01-18
Create a tool to migrate CAPI1 cores to CAPI2


olofk:
Lots of cores need to eventually be converted. Having a tool to do this automatically could be helpful

imphil:
Will be fixed by #231

PR #231, opened 2018-11-03
Implement core converter

Oxore:
I saw #188 issue and decided to try to implement a CAPI1 to CAPI2 core converter. It still needs testing, though, that I can't fully provide.

There are some questions and things, that, I think, worth noticing.

1. Additional 'conv' subparser added to `fusesoc/main.py` to use core converter. If you think, that there is a better way to introduce converter, I would like to know about it.

2. Dependencies of some related file sets become shared and redundant. For example, `verilog_tb_src_files` and `verilog_tb_private_src_files` will have the same copied depend list.

3. `rivierapro` is not listed in the `SimulatorList` class in `fusesoc/capi1/section.py`. Is this because nobody had a chance to test it with fusesoc yet?

4. `verilator` treated as a simulator in the `SimulatorList` class in `fusesoc/capi1/section.py`, but at the same time it is treated as a synthesizer in `_get_flow` method of the `Core` class in fusesoc/capi1/core.py. So I made this converter definitely treat `verilator` as a synthesizer.

5. CAPI1 does not treat the `vhdl` section as a `fileset` like it does with verilog, for instance, and do not put it to `Core.file_sets`. Is it OK? Does it worth worrying about? See the `_collect_filesets` method of the `Core` class in `fusesoc/capi1/core.py`.

6. Yaml dumper does not indent file lists, may be a little uncomfortable. Yaml output also is rough and in alphabetical order, of course. Do you think a beautifier is necessary?

7. Scripts migration is not so simple: full file path to script must be specified in the CAPI2 core file. So I made this in a similar way as it appears to be in a generated `<corename>.eda.yml`. Every script is also added to it's own file set and mentioned in all targets: `default`, `sim` and `synth`.

8. The only tool I've properly tested is Quartus and the only simulator is Icarus Verilog. For example, `de0_nano` core is converted fine, according to diff of produced `de0_nano.eda.yml` file. Other tools related conversion implemented in a similar way and hence there is a chance, that they will be converted without errors.


olofk:
Fantastic! Thanks for working on this. I currently have a big backlog of things I need to take care of, so I can't unfortunately give a proper review right now, but I'll answer the questions you listed at least

>     1. Additional 'conv' subparser added to `fusesoc/main.py` to use core converter. If you think, that there is a better way to introduce converter, I would like to know about it.
> 

I  want a new subcommand called `core` and this to be called `upgrade`. (I'm planning to change `list-cores` to `core list` and `core-info` to `core info` and eventually add a couple of other core commands)

>     2. Dependencies of some related file sets become shared and redundant. For example, `verilog_tb_src_files` and `verilog_tb_private_src_files` will have the same copied depend list.
> 

It's a bit ugly, but I think that's the way it has to be done. Hopefully the old `verilog` section will die out soon

>     3. `rivierapro` is not listed in the `SimulatorList` class in `fusesoc/capi1/section.py`. Is this because nobody had a chance to test it with fusesoc yet?
> 

I must have missed that in a refactoring at some point. I implemented support for rivierapro around two years ago, but have not had access to a license since then. Haven't heard any complaints, so it might be very little tested. Do what's easiest for you. I don't mind pretending Riviera Pro support is only available in CAPI2

>     4. `verilator` treated as a simulator in the `SimulatorList` class in `fusesoc/capi1/section.py`, but at the same time it is treated as a synthesizer in `_get_flow` method of the `Core` class in fusesoc/capi1/core.py. So I made this converter definitely treat `verilator` as a synthesizer.
> 

Yeah. That has always been a bit of a mess, and a clear indication that dividing tools into simulators and synthesis tools didn't really work out as I had hoped originally. Just do what's easiest here

>     5. CAPI1 does not treat the `vhdl` section as a `fileset` like it does with verilog, for instance, and do not put it to `Core.file_sets`. Is it OK? Does it worth worrying about? See the `_collect_filesets` method of the `Core` class in `fusesoc/capi1/core.py`.
> 

Ignore the VHDL section. It has been deprecated for a long time (although I just noticed that I had not mentioned this in https://github.com/olofk/fusesoc/blob/master/doc/migrations.adoc)

>     6. Yaml dumper does not indent file lists, may be a little uncomfortable. Yaml output also is rough and in alphabetical order, of course. Do you think a beautifier is necessary?
> 

No, that's fine. It's important though that files within a fileset appears in the correct order. Otherwise it might cause failures with VHDL

>     7. Scripts migration is not so simple: full file path to script must be specified in the CAPI2 core file. So I made this in a similar way as it appears to be in a generated `<corename>.eda.yml`. Every script is also added to it's own file set and mentioned in all targets: `default`, `sim` and `synth`.
> 

That's fine. I don't think the scripts section was used too much, so if someone has problems, it's a good reason to fix this up manually afterwards

>     8. The only tool I've properly tested is Quartus and the only simulator is Icarus Verilog. For example, `de0_nano` core is converted fine, according to diff of produced `de0_nano.eda.yml` file. Other tools related conversion implemented in a similar way and hence there is a chance, that they will be converted without errors.

I will take a closer look at the generated output for all the tools when I find time to do a proper review.

Again, thanks for doing this. It's highly appreciated

Oxore:
I'm glad to help! I will be in touch if you need me to rewrite or fix something after your review.

imphil:
Thanks @Oxore for your work, and sorry for letting it sit around for that long. I've just updated this PR to fix some smaller issues and renamed the command to `migrate-capi1-to-capi2` to indicate that it is purely a transitional command that will go away in the FuseSoC 2.0 release.

To convert a full directory of core files, use the following command:

```
find your_coredir -iname '*.core' -exec fusesoc migrate-capi1-to-capi2 --nowarn --inplace {} \;
```

@olofk I'd suggest merging this as-is mostly to give people a starting point for converting their CAPI1 core to CAPI2. It's not perfect, and unlikely that it ever will be, but it's still better than doing all the conversation by hand.

olofk:
Many thanks @Oxore and @imphil for working on this. Let's bring it in. I'm in the process of converting a bunch of cores so this will be perfect!

olofk:
FYI, I just converted my first core and running the regression tests works just fine. Excellent work!

PR #234, opened 2018-11-22
Spyglass: Fix parameter passing

imphil:
Spyglass requires parameter to always be in a dotted (hierarchical)
syntax, which is different from most other tools (e.g. Vivado). To have
a consistent interface prefix the parameter name with the toplevel
module name before passing to Spyglass.

### DIFF
--- a/edalize/spyglass.py
+++ b/edalize/spyglass.py
@@ -1,5 +1,6 @@
 import logging
 import re
+from collections import OrderedDict
 
 from edalize.edatool import Edatool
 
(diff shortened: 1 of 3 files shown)


imphil:
2nd try without failing tests (at least locally)

olofk:
I planned to push this to edalize using the method you outlined in the other spyglass PR. Unfortunately there are conflicts now since I pushed the previous PR so I'm a bit anxious about messing up your changes.

I also just kicked out edalize from fusesoc, so could you please submit a rebased version against edalize instead and I'll promise to merge it quickly (less stuff on my table now)

olofk:
FYI, icarus uses the same format for parameters. I just looked at the icarus source and noticed I'm cheating a bit since the icarus implementation will not work with the (quite rare) situation of multiple top-levels. I see you don't handle that either, but let's ignore that until someone complains :)

olofk:
I took another look and it was all straight-forward to rebase, so I took the liberty of applying this to edalize myself https://github.com/olofk/edalize/commit/ee0afd6b586f369be24acd529381d0b2d1891ad9

Please shout if I broke something

PR #241, opened 2019-02-13
Add ability to run fusesoc without a backend tool for file generation…

benreynwar:
… purposes.

Adding a special tool called "none" that causes the '"run" command to exit after the initial generation stage.
This is useful if you just want to use fusesoc for dependency management and file generation sometimes.


olofk:
It's a very simple fix so perhaps I'm overthinking things here, but I fear that hijacking the tool flag would cause problems for targets that e.g. generates different things depending on which tool is selected. An alternative solution would be to have a dedicated `export` command (or option to the `run` command). I have been thinking about such a command already, and even allow it to accept multiple tools so that you can create project files for e.g. both modelsim and quartus at the same time.

I can't find any *real* issues with your proposed change, but just let me think about it to see if it might clash with something. Otherwise I'm fine with it

benreynwar:
This is unnecessary because you can run

`fusesoc run --target=$target --tool=$tool --setup $core`

to generate the tool files without running the tool.
(olofk pointed this out on the gitter channel)

Issue #238, opened 2019-01-18
"fusesoc gen" should give better error message


benreynwar:
Currently it gives the general parser help message.  It should give the help message for the 'gen' subparser.

(Hi Olof.  I'm mostly adding this to remind myself to look at it when I'm not at work.  Not trying to add more things to your plate :) )

olofk:
And look at it you did. Thanks for the fix

PR #244, opened 2019-02-13
Show subparser error message when subparser was used but no func was …

benreynwar:
…set.

Fixes issue #238


olofk:
Love it! I've been annoyed by this too :) Picked and pushed. Thanks!

PR #245, opened 2019-02-14
Fix bug in generator core pos logic.

benreynwar:
Currently the default is for a generator snippet to be placed at the end.  I think this is just a typo.  It would make more sense for the default to be just append it to the list of central snippets.


olofk:
Yes, you're right. It took me unreasonably long time to add this simple feature and I managed to mess it up in the end anyway :) Picked and pushed. Thanks!

PR #249, opened 2019-03-14
Avoid yaml.load deprecation warning

GCHQDeveloper560:
A fresh FuseSoC build gives the following warnings:
```
[...]/site-packages/ipyxact/ipyxact.py:178: YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated, as the default Loader is unsafe. Please read https://msg.pyyaml.org/load for full details.
  _generate_classes(yaml.load(ipxact_yaml.description))
[...]/site-packages/fusesoc/capi2/core.py:744: YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated, as the default Loader is unsafe. Please read https://msg.pyyaml.org/load for full details.
  capi2_data = yaml.load(description)
[...]/site-packages/fusesoc/capi2/core.py:116: YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated, as the default Loader is unsafe. Please read https://msg.pyyaml.org/load for full details.
  _root = Root(yaml.load(open(core_file)))
```

This handles the two FuseSoC warnings by adding the requested `Loader` argument.

### DIFF
--- a/fusesoc/capi2/core.py
+++ b/fusesoc/capi2/core.py
@@ -741,7 +741,7 @@ def _generate_classes(j, base_class):
         generatedClass = type(cls, (base_class,), class_members)
         globals()[generatedClass.__name__] = generatedClass
 
-capi2_data = yaml.load(description, Loader=yaml.FullLoader)
+capi2_data = yaml.safe_load(description)
 
 for backend in get_edatools():
     backend_name = backend.__name__
--- a/fusesoc/capi2/core.py
+++ b/fusesoc/capi2/core.py
@@ -113,7 +113,7 @@ class Core:
         self.core_root = os.path.dirname(core_file)
 
         try:
-            _root = Root(yaml.load(open(core_file)))
+            _root = Root(yaml.load(open(core_file), Loader=yaml.FullLoader))
         except KeyError as e:
             raise SyntaxError("Unknown item {}".format(e))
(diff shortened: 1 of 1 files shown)


olofk:
This is the right thing to do, but I took a quick look now and it seems that FullLoader isn't available on my yaml installation (3.13). I would like to investigate this a bit to see if there's a way to support both older and newer pyyaml version (e.g. use Loader=yaml.Loader instead).

GCHQDeveloper560:
Sounds good. Sorry I didn't think to check backward compatibility.

It looks like the older versions include SafeLoader or the shorter safe_load. I don't think edalize is doing anything like loading classes from YAML that isn't supported by that loader, so that could be a way to go as well.

olofk:
I think you're right that using safe_load is the better solution here. Looking at the list of open bugs for pyyaml 5.1 I get the feeling it might be better to not require that version as it looks a bit unstable. Would you like to make a new patch that uses yaml.safe_load instead or do you want me to make the changes? Both are fine with me

GCHQDeveloper560:
I've pushed the safe_load version. I'm getting the same warning for ipyxact, so olofk/ipyxact#16 applies the same fix.

olofk:
Thanks! Picked and pushed now. I took the liberty to squash these into a single commit and also convert the other uses of yaml.load. Hope this is ok with you

PR #253, opened 2019-03-18
Documentation: use RTD theme

imphil:
The Alabaster theme isn't helping readability by having too little
structure for the eye to see where the navigation and where the content
is. The RTD theme is what people are used to, and it's well designed for
documentation. Use that if available.


olofk:
Wow! That looks much better. Picked and pushed! Thanks

PR #254, opened 2019-03-19
Fix link in README.rst to point to up-to-date documentation

jschiefer:
The README.rst on the repo points to an outdated version of the docs. It would be nice if it pointed to the [readthedocs.io site](https://fusesoc.readthedocs.io/en/latest/index.html) instead. 

Also, adding a link to edalize makes it easier to figure out what that is.


olofk:
This sounds like a very reasonable change. Picked and pushed. Thanks!

PR #259, opened 2019-05-06
Add a aresetn signal to ensure consistent simulation between tools

hmaarrfk:
The xsim simulator doesn't start clock and reset signals in the same state as icarus.

The original example, when simulated with xsim thinks that the very first beat happens after `999_980_000` timescales and not  `1_000_000_000`.

I think adding the reset logic helps define some "undefined initial starting state" which helps the simulator.

Feel free to rename the signal as you wish.

### DIFF
--- a/doc/tutorials/2-creating_a_core.md
+++ b/doc/tutorials/2-creating_a_core.md
@@ -39,13 +39,14 @@ module pps
   #(parameter clk_freq_hz = 50_000_000)
    (input  clk,
     input  aresetn,
-    output reg pps_o = 1'b0);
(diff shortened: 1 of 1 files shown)


imphil:
Thanks for your contribution @hmaarrfk! (Un)fortunately, this tutorial is now gone and replaced with a new user guide. Let me know if you find more issues with the current new code!

hmaarrfk:
Its all good.

Still trying to wrap my head around using your tooling system.

I think one day I'll find that SOC tools are as good as software tools. I think what you've created is a step toward that!

PR #271, opened 2019-05-10
Adding a --system-name flag

Jbalkind:
We talked about this at some point and I couldn't remember which of us had the patch. You might want something more elaborate but how about this?


olofk:
Thanks for this. Still can't believe how easy that was. Picked and pushed

PR #276, opened 2019-06-07
Init default library location

towoe:
Use the 'fusesoc' subdirectory for the user output and also for the
directory exists check.

As the creation of the path for the library is done at a different point, this should not change the behavior for the location of the default library folder, but gives the correct default location for the user.


olofk:
Many thanks. This is absolutely correct. Picked and pushed

PR #281, opened 2019-06-19
Add file-specific logical names

mscoolnerd:
This change makes it so that individual files can be in different libraries, similar to how the "is_include_file" member works.


olofk:
Whoops. I thought that was already implemented but apparently not. Thanks for the patch. I took the liberty to fix the typo in the commit header and add a test that uses this new feature. Hope that's fine or let me know otherwise

PR #288, opened 2019-07-24
Add warnings for deprecated sim and build commands.

m-kru:
These commands were intended to be used with CAPI1 format.


olofk:
Thanks

PR #293, opened 2019-07-29
Allow generator data to be supplied to the constructor

GCHQDeveloper560:
It seems useful to allow a generator to be run as a Python module directly from a script rather than having to be run as a separate process. For example, a script could drive fusesoc/edalize and try a core with several different package settings or clock constraints.

I think a script would rather supply the configuration data directly in a dictionary rather than going through YAML, so this change allows a script to supply the data to the generator's constructor. For backward compatibility the default case loads YAML from a file on the command line.

The generator inheriting from this base class would also need to support this usage. I'll follow up with a change to the template generator that seems to work.

### DIFF
--- a/fusesoc/capi2/generator.py
+++ b/fusesoc/capi2/generator.py
@@ -5,16 +5,17 @@ class Generator(object):
     filesets   = {}
     parameters = {}
     targets    = {}
-    def __init__(self):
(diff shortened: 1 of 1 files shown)


olofk:
This is great and will help other projects as well that wants to reuse the generator infrastructure. Thanks. Picked and pushed

PR #307, opened 2019-10-23
Pass on verbosity to edalize

imphil:
Log verbosity preferences (e.g. passed through --verbose) are applied
only to fusesoc, not to edalize currently.

That's slightly unexpected since edalize is "invisible" to the end user
and integrated into the flow. This simple change makes the logging
settings apply to edalize as well.


olofk:
Very nice! Had no idea you could do that. Picked and pushed

PR #318, opened 2019-11-25
generators-library

a-gibson:
Allow `logical_name` to be specified when adding files to a generator core file.


olofk:
Sure thing. Thanks for the patch!

PR #320, opened 2019-12-03
Fix VLNV description

towoe:


olofk:
Whoops :) Thanks for that

PR #321, opened 2019-12-18
Rework documentation, part 1

imphil:
First part of a documentation rework. Move a couple files around to structure them in a better way. 

There's more to come (including new content), starting with this one to keep a sane stack of patches.


olofk:
Fantastic! Many thanks for doing this. The new structure makes so much more sense

PR #323, opened 2019-12-23
Add more content to documentation

imphil:
A random selection of new documentation content. Not all of it is fully polished, but it should be good enough for a starting point and attract more contributions in the future (including from me; I'll look at the API docs next).

### DIFF
--- a/doc/source/dev/devsetup.rst
+++ b/doc/source/dev/devsetup.rst
@@ -1,0 +1,35 @@
+Development Setup
+=================
+
+.. note::
+
+   To make changes to a backend, e.g. to the way a simulator or synthesis tool is called, you need to modify edalize, not fusesoc.
+   Edalize is a separate project, see https://github.com/olofk/edalize for more information.
+
+Linux
+-----
+
+.. note::
+
+   If you have already installed FuseSoC, remove it first using ``pip3 uninstall fusesoc``.
+
+The FuseSoC source code is maintained in a git repository hosted on GitHub.
+To improve FuseSoC itself, or to test the latest unreleased version, it is necessary to clone the git repository first.
+
+.. code-block:: bash
+
+   cd your/preferred/source/directory
+   git clone https://github.com/olofk/fusesoc
+
+FuseSoC then needs to be installed in editable or development mode.
+In this mode, the ``fusesoc`` command is linked to the source directory, and changes made to the source code are immediately visible when calling ``fusesoc``.
+
+.. code-block:: bash
+
+   cd fusesoc
+   # Use the ``--user`` flag to perform an installation for the current user;
+   # remove this flag for a system-wide installation.
+   pip3 install --editable --user .
+
+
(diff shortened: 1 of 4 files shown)


wallento:
When there is one thing that FuseSoC doesn't need, it's more documentation..

Great work, @imphil, thanks

doc/source/dev/devsetup.rst
```diff
@@ -0,0 +1,35 @@
+Development Setup
+=================
+
+.. note::
+
+   To make changes to a backend, e.g. to the way a simulator or synthesis tool is called, you need to edalize, not fusesoc.
```
olofk: Lost a verb there between to and edalize. update, edit, change?

doc/source/dev/devsetup.rst
```diff
@@ -0,0 +1,35 @@
...
+   Edalize is a separate project, see https://github.com/olofk/edalize for more information.
+
+Linux
+-----
+
+.. note::
+
+   If you have already installed FuseSoC, remove it first using ``pip3 uninstall fusesoc``.
+
```
olofk: Why? Because pip won't update by default? We could suggest install -U instead. Same end result though. Or is it when you want to install with -e for development. Don't think you need to uninstall in that case. Your way works fine. Just want to check I didn't miss anything

doc/source/dev/testing.rst
```diff
@@ -0,0 +1,20 @@
...
+
+Running tests
+-------------
+
+.. code-block::
+
+   cd fusesoc/source/directory
+
+   # run all tests
+   pytest
```
olofk: pytest doesn't work for me anymore and I have to run py.test instead. Not sure if this is just my system or a general thing. Also wonder if we have the same issue as has been found with edalize so that we need to run with `PYTHONPATH=. py.test` to ensure it picks up the local version

imphil: - py.test is the old binary name, pytest is the newer one (see https://stackoverflow.com/questions/39495429/py-test-vs-pytest-command). `python3 -m pytest` should always work, I've switched to that.
- PYTHONPATH is only needed if you didn't install fuseosc before calling pytest. I've edited the document to explicitly include this step.

doc/source/dev/devsetup.rst
imphil: This ensures that only a single installation remains after executing these steps.

The problem are per-user vs global installations. pip uninstall will remove all existing installations, both global and per-user ones. pip install -U is less reliable. Think of the following scenario:

```
pip3 install -U --user fusesoc
sudo pip3 install -U fusesoc

# now you have two installations, one in /usr/local, and one in ~/.local
```

I've seen a fair amount of people confused by that behavior, and pip uninstall is not much harder but slightly more reliable.

imphil:
Thanks for your comments @olofk! PR updated.

olofk:
All good. Your comments make sense. Picked and pushed. Thanks!

PR #327, opened 2019-12-24
Mark fusesoc as stable in setuptools/pypi

imphil:
FuseSoC is developed in a stable manner and used daily in production
environments, change trove classifier to reflect this fact.


olofk:
Woohoo. We're out of beta :)

imphil:
Out of alpha actually!

PR #328, opened 2019-12-24
Add python_requires to setup.py

imphil:
This feature enables pip to prevent the installation if Python is too
old.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -43,4 +43,6 @@ setup(
         'pyyaml',
         'simplesat>=0.8.0',
     ],
+    # Supported Python versions: 2.7 or 3.5+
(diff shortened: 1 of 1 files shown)


imphil:
BTW the syntax looks ugly, but it's the only way of doing it. (I didn't believe it either in cocotb a couple weeks back.)

olofk:
Yikes. That looks awful. Thought semver had worked these things out. Oh well. It's a good patch. Thanks

Fatsie:
I though it could be done with '~=2.7, ~=3.5'.
See https://www.python.org/dev/peps/pep-0440/#compatible-release

imphil:
@Fatsie that expression would select nothing: "The comma (",") is equivalent to a logical and operator: a candidate version must match all given version clauses in order to match the specifier as a whole."

PR #333, opened 2020-01-06
Run black on the whole source tree

imphil:
Run the black formatter to reformat the whole source tree according to
the black coding style
(https://black.readthedocs.io/en/stable/the_black_code_style.html).


imphil:
See https://github.com/olofk/fusesoc/issues/334 for a discussion on this PR.

PR #336, opened 2020-01-09
Add pre-commit hooks to run lint and formatters

imphil:
pre-commit is a small helper tool which makes it easy to install linters, formatters and similar tooling. Document how to use it, and enable it in CI.


imphil:
This PR is based on #335 and needs to be rebased once the other PR goes in.

PR #339, opened 2020-01-15
Remove unused edalize test

imphil:
This test seems to left over in the fusesoc repository when edalize was
factored out. Remove it.


olofk:
Cool stuff. Commits that remove stuff are the best commits

Issue #315, opened 2019-11-21
Supporting dependency ranges in CAPI file filesets


GCHQDeveloper992:
I was thinking it might be beneficial to support a dependency range in CAPI files to cut down on repetitive typing. For example, I have a number of cores that are semantically versioned so I often want to include the range of versions between one major release and the next, e.g.

```yaml
CAPI=2:
name: somevendor:somelibrary:somecore:1.0.0

filesets:
  rtl:
    depend:
      - ">=somevendor:somelibrary:anothercore:1.0.0"
      - "<somevendor:somelibrary:anothercore:2.0.0"
     ....more capi file....
```

For some cores that have lots of dependencies, this gets quite repetitive. I was wondering if the following syntax might be nicer:

```yaml
filesets:
  rtl:
    depend:
      - ">=<somevendor:somelibrary:anothercore:1.0.0/2.0.0"
     ....more capi file....
```

I believe with a small modification to coremanager.py _parse_depend function we could identify the new >=< relation operator, and automatically append two strings to the deps array - one for the >= relation for the 1.0.0 version before the '/' and another for the < relation for the 2.0.0 after the slash.

I could try and put together a quick demo branch of this behaviour to see if that's all that would be involved.

Does that sound like a useful feature for more than just me? Or have I missed a capability that already exists?

GCHQDeveloper992:
Having looked at making a quick test branch, I think it might be a bit trickier than I originally thought. Putting in core manager duplicates a bunch of parsing logic from the Vlnv class. I also tried having an @classmethod on Vlnv that could return a list of Vlnvs (so that the Vlnv constructor could be called multiple times) but this causes trouble with handling the return values.

I'll give this another go later on to see if I can find something clearer that achieves the result.

GCHQDeveloper992:
I've taken another shot at something that didn't introduce huge numbers of changes. You can see it in my fork [here](https://github.com/GCHQDeveloper992/fusesoc/commit/61f5281946fd3f5475601e64608d2c7ac50fe4f2)

Not sure if this is the neatest implementation though.

olofk:
This one slipped between the cracks. Sorry about that. I think the intention is good, but I would suggest using some existing syntax for ranges. The first that comes to mind is the [NPM semver](https://semver.npmjs.com/) which I think is the same that python's own dependency manager uses (or probably something slightly incompatible, just to make life more annoying). Or perhaps that's just more complicated than it needs to be

[...]

PR #354, opened 2020-02-13
Add dependency ranges based on NPM syntax

GCHQDeveloper992:
As discussed in issue #315 it would be useful to be able to specify ranges of version numbers in a single depend statement, to minimise the amount of typing in core files that make use of semantic versions. 

As suggested by @olofk this PR implements the ^ and ~ syntax from https://semver.npmjs.com which allows you to specify ranges of versions relative to the specified major version and minor version respectively. For example "^::test:1.2.3" would search for a test core between from 1.2.3 up to 2.0.0, whilst "~::test:1.2.3" would search from 1.2.3 up to 1.3.0.

This tries to minimise the amount of code that is changed to support this, though does introduce a way of asking a VLNV to describe itself in terms of "simple" VLNVs (i.e. VLNVs that only use an operator and a single version). This helps minimise the amount of new things the core manager needs to know about and means that we don't have to teach the simplesat solver about new operators.

As a test, I've used the following core file and demonstrated that I successfully pull different versions of stream_utils for each target.

```
CAPI=2:

name: test:testing:testcomp:0.0.0

filesets:
  testmajor:
    depend:
      - ^::stream_utils:1.1
    files:
      - test/test.vhd
    file_type: vhdlSource-2008
    logical_name: test

  testminor:
    depend:
      - ~::stream_utils:1.1
    files:
      - test/test.vhd
    file_type: vhdlSource-2008
    logical_name: test

targets:
  major:
    filesets: [testmajor]
    toplevel: test
    default_tool: quartus
    tools:
      quartus:
        device: 10AX115N3F45E2SG
        family: Arria10

  minor:
    filesets: [testminor]
    toplevel: test
    default_tool: quartus
    tools:
      quartus:
        device: 10AX115N3F45E2SG
        family: Arria10
```


GCHQDeveloper992:
Looks like the failure from Appveyor might be due to it pulling in Jinja2 3.0.0a. 

I noticed that the master branch of edalize now ties Jinja2 to 2.8 through 2.11.0 so I assume this is either the same or similar to the bug imphil raised?

olofk:
Hi,

I'm just in the process of making a new FuseSoC release. Was hoping to add your work as well. In the process, there will soon be a new edalize release so let's just wait for that to solve the jinja issues.

Is the code good to go otherwise? Just about to start reviewing it

GCHQDeveloper992:
I think the code is good to go, unless a review flags up that a better approach could be taken. Thanks for taking a look

olofk:
Thanks. It looks fine to me so I will rebase your patches and squash it into one commit with the topic of the PR if you are fine with that

I also have another question tangentially related to this. We are looking into the possibility of a relicensing of FuseSoC and for that we would need the blessing of all current code contributors before moving ahead. Would you be willing to have your FuseSoC contributions relicensed to a BSD license. The same question goes to @GCHQDeveloper560 for whom I also lack any other contact information

GCHQDeveloper992:
Squash and rebase sounds good to me.

On the topic of licensing, I'd be happy for my contributions to be under the BSD license. I'll flag this comment up to @GCHQDeveloper560 and have them comment here for you.

olofk:
Perfect! Thank you. I ended up doing a bit of manual rebase. Could you please just do a sanity check on https://github.com/olofk/fusesoc/tree/deprange before I push to master, so that I didn't mess up anything?

GCHQDeveloper992:
All looks good to me 👍

GCHQDeveloper560:
I am also okay with my contributions being BSD licensed. I am curious about the reasoning behind the license change if you're able to share more.

olofk:
@GCHQDeveloper992 Great! Pushed now. Thanks for your contributions

@GCHQDeveloper560 Thanks. The main reason is that I have gotten the question from some larger companies that want to integrate FuseSoC (or parts of FuseSoC) as a library inside their tools. In the cases I'm aware of, these tools are BSD-licensed and integrating a GPLv3 module would need them to relicence their code base for license compatibility.

Issue #357, opened 2020-02-18
FileExistError exception thrown on repeated runs


sriyerg:
When running sims repeatedly without cleaning the build area, I see the following exception thrown:

```console
Traceback (most recent call last):
  File "/usr/local/google/home/sriyer/.local/bin/fusesoc", line 10, in <module>
    sys.exit(main())
  File "/usr/local/google/home/sriyer/.local/lib/python3.7/site-packages/fusesoc/main.py", line 583, in main
    args.func(cm, args)
  File "/usr/local/google/home/sriyer/.local/lib/python3.7/site-packages/fusesoc/main.py", line 268, in run
    flags, args.system_name, args.system, args.backendargs, args.build_root)
  File "/usr/local/google/home/sriyer/.local/lib/python3.7/site-packages/fusesoc/main.py", line 316, in run_backend
    system_name=system_name)
  File "/usr/local/google/home/sriyer/.local/lib/python3.7/site-packages/fusesoc/edalizer.py", line 53, in __init__
    core.export(files_root, _flags)
  File "/usr/local/google/home/sriyer/.local/lib/python3.7/site-packages/fusesoc/capi2/core.py", line 183, in export
    os.makedirs(os.path.join(dst_dir, d))
  File "/usr/lib/python3.7/os.py", line 221, in makedirs
    mkdir(name, mode)
FileExistsError: [Errno 17] File exists: 'scratch/top_earlgrey.sim.vcs/top_sanity/default/src/lowrisc_systems_top_earlgrey_0.1/../ip/alert_handler/rtl'
```
A quick fix for this is to simply set `exist_ok=True` flag to `os.makedirs()` in `fusesoc/capi2/core.py` :: line 183.

sriyerg:
Fixed this issue by better reorganizing the sources in the core files (ensuring core files do not include sources from a different directory tree.)

PR #358, opened 2020-02-18
Fix for #357

sriyerg:
When running sims multiple times, directories may exist from previous
runs. This fix prevents `FileExistError` from being thrown when fusesoc
attempts to create directory trees which may already exist.

Signed-off-by: Srikrishna Iyer <sriyer@google.com>

### DIFF
--- a/fusesoc/capi2/core.py
+++ b/fusesoc/capi2/core.py
@@ -205,7 +205,7 @@ class Core:
         dirs = list(set(map(os.path.dirname, src_files)))
         for d in dirs:
             if not os.path.exists(os.path.join(dst_dir, d)):
-                os.makedirs(os.path.join(dst_dir, d))
+                os.makedirs(os.path.join(dst_dir, d), exist_ok=True)
 
         for f in src_files:
             if not os.path.isabs(f):


imphil:
The root cause for that is the OpenTitan use of including files with  which are above the core file (`../file.sv`). This is unsupported in fusesoc. Your patch works around that, but it's papering over the issue that we need to solve in OpenTitan, because we'll run into it again in other scenarios. 

I'm still OK with taking this patch, but you need to modify it in a way that works with Python 2.7 for CI to pass.

sriyerg:
> The root cause for that is the OpenTitan use of including files with which are above the core file (`../file.sv`). This is unsupported in fusesoc. Your patch works around that, but it's papering over the issue that we need to solve in OpenTitan, because we'll run into it again in other scenarios.
> 
> I'm still OK with taking this patch, but you need to modify it in a way that works with Python 2.7 for CI to pass.

Makes sense. I re-org'ed the sources and all the core files in OpenTitan to resolve this in PR [#1570](https://github.com/lowRISC/opentitan/pull/1570). I will close this PR since it is not needed.

sriyerg:
The sources in Opentitan fusesoc core files are now better organized. With that, this exception is no longer thrown on repeated runs.

PR #362, opened 2020-03-04
Code cleanups

imphil:
This contains a couple of code cleanups and small enhancements, see the individual commit messages for full details.

This is in preparation for a larger chunk of work on generators, which will come in the next couple of days. I want to get this out of the way first, though.

### DIFF
--- a/fusesoc/vlnv.py
+++ b/fusesoc/vlnv.py
@@ -92,17 +92,12 @@ class Vlnv(object):
         self.sanitized_name = str(self).lstrip(":").replace(":", "_")
 
     def __str__(self):
-        if self.relation == "=":
-            relation = ""
-        else:
-            relation = self.relation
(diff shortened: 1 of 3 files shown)


fusesoc/vlnv.py
```diff
@@ -92,17 +92,12 @@ def _is_version(s):
         self.sanitized_name = str(self).lstrip(":").replace(":", "_")
 
     def __str__(self):
-        if self.relation == "=":
-            relation = ""
-        else:
-            relation = self.relation
```
olofk: It looks to me like this would no longer print e.g. `>=vendor:library:name:version` but maybe this isn't used anywhere?

olofk: Ok, looked at the code now. Yeah, it's dead.

PR #364, opened 2020-03-04
More speed improvements

imphil:
In preparation for some generator work, here are two more speed improvements for fusesoc.

* Use a faster YAML parser if possible. This should be quite visible in projects with a significant amount of CAPI2 core files, and a large EDAM file.
* Cache the solved dependency tree. This won't be visible at this point, since the solving only takes place once. But as part of my work on generators, the solving will need to happen more than once, and caching helps significantly in this case.

Again, see the individual commit messages for details.

### DIFF
--- a/fusesoc/capi2/core.py
+++ b/fusesoc/capi2/core.py
@@ -147,7 +147,7 @@ class Core:
         self.core_root = os.path.dirname(self.core_file)
 
         try:
-            _root = Root(yaml.safe_load(open(self.core_file)))
+            _root = Root(utils.yaml_fread(self.core_file))
         except KeyError as e:
             raise SyntaxError("Unknown item {}".format(e))
         except (yaml.scanner.ScannerError, yaml.constructor.ConstructorError) as e:
@@ -855,7 +855,7 @@ def _generate_classes(j, base_class):
         globals()[generatedClass.__name__] = generatedClass
 
 
-capi2_data = yaml.safe_load(description)
+capi2_data = utils.yaml_read(description)
 
 for backend in get_edatools():
     backend_name = backend.__name__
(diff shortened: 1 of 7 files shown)


GCHQDeveloper560:
PyYAML gave us understandably nasty warnings in the past about using `load` vs `safe_load`. Are you planning to load stuff from YAML not supported by `safe_load`? It looks like the `safe_load` calls you replaced don't support the `Loader` argument, but I think the equivalent would be to use `SafeLoader`/`CSafeLoader` and `SafeDumper`/`CSafeDumper`.

imphil:
There's no CSafeLoader (that's always "safe"). But you're right, I should be using SafeLoader instead of Loader. I'll update the patch. `safe_load()` is just a shortcut to `load(..., Loader=SafeLoader)`

GCHQDeveloper560:
I'm not sure what you mean by `CSafeLoader` not always being safe. The [docs](https://pyyaml.org/wiki/PyYAMLDocumentation) make it sound like `CSafeLoader` is equivalent to `SafeLoader`:

> SafeLoader(stream) supports only standard YAML tags and thus it does not construct class instances and probably safe to use with documents received from an untrusted source. The functions safe_load and safe_load_all use SafeLoader to parse a stream.
...
> CLoader, CSafeLoader, CBaseLoader are versions of the above classes written in C using the LibYAML library.

In my quick test I was able to do something simple with `CSafeLoader` but didn't try to make sure it refused something too complex.

imphil:
Ah, good catch! I was looking through the docs before but didn't find CSafeLoader. So I'll give that a try.

imphil:
I've updated the patch containing the loader/dumper code to now use CSafeLaoder/CSafeDumper if possible, keeping the behavior we had before.

olofk:
Looks great. Thanks. Great to get some caching. IIRC, the dependency tree is always calculated twice but I've been too lazy to fix that and no one complained until now :)

Feel free to commit

PR #368, opened 2020-03-10
Fix potential race condition to create cache root

imphil:
Avoid a race condition between checking for the cache root and actually
creating it.


imphil:
This PR intentionally doesn't work in Python 2.7 any more; let's not merge it before we explicitly state that by updating the metadata.

olofk:
You have my blessing to remove the py2.7 support now that FuseSoC 1.10 has been released

PR #370, opened 2020-03-10
Fix Appveyor CI

imphil:
Ensure that we use a recent enough version of pip and setuptools in CI to parse all pip metadata.


imphil:
Seems to do the trick.

PR #379, opened 2020-03-18
Add command to draw a dependency graph

### LINKED ISSUE #98: Plot the dependency graph with graphviz
The idea is to plot a dependency graph for the given core/system

imphil:
The new command `fusesoc dep graph` creates a GraphViz .dot file
listing all dependencies for a given core with given flags. This file
can be either manually converted into an image using the `dot` tool
(part of the GraphViz package), or automatically converted with FuseSoC
through the `--output-format` command line option.

Full documentation:

```
$ fusesoc dep graph --help
usage: fusesoc dep graph [-h] [--target TARGET] [--tool TOOL] [--output-format OUTPUT_FORMAT] core

positional arguments:
  core                  Core to build the dependency graph for.

optional arguments:
  -h, --help            show this help message and exit
  --target TARGET       Target to build the dependency graph for.
  --tool TOOL           Tool to build the dependency graph for.
  --output-format OUTPUT_FORMAT
                        Convert output into desired format (requires dot). All dot-supported formats can be used, run 'dot -T?' for a complete list.
```

Based on inital work by Stefan Wallentowitz in #99

Fixes #98

### DIFF
--- a/fusesoc/coremanager.py
+++ b/fusesoc/coremanager.py
@@ -12,6 +12,8 @@ from simplesat.request import Request
 
 from fusesoc.core import Core
 from fusesoc.librarymanager import LibraryManager
+from fusesoc.vlnv import Vlnv
+from fusesoc.utils import merge_dict
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


imphil:
Hm, thinking about it, I might need to rework this PR to take cores produced as the result of a generator into account. Please wait a bit before reviewing.

imphil:
So the question becomes:
- If we want to go with an approach like in this PR, we'll only get dependencies without generators-produced dependencies.
- If we want to include generator-produced dependencies, we need a full build environment. I did a prototype where I always produce a *.dot file next to the .eda.yml file as part of the `run` target. We get less configurability this way, but also avoid having to add another command line option and all the like.

@wallento @olofk What approach sounds best to you? I'd go for the second option.

olofk:
On one hand, I'm still sceptical towards allowing generated cores to have dependencies (especially with the current dependency manager), so for that reason I would be ok with option 1.

At this point however, I'm a bit reluctant to shove even more options and code into main.py, so generating the dot file together with the EDAM file seems like a good solution for now. As long as you don't have too many generators, there shouldn't be much overhead to generate the build tree anyway

olofk:
Needs a rebase though

imphil:
In OpenTitan we ran the dot file generation as part of the build process for some time now, and didn't see significant performance problems. So I'll go for that approach and will submit a PR once the generator-dependency-thing discussion is over.

Issue #363, opened 2020-03-04
Rewrite of argument parsing


olofk:
I'm working on rewriting the argument parsing. Since this is a fairly long process and involves at least @imphil and @kgugala and probably others I wanted to write it down here so we can discuss it in public.

The maon goal is to get rid of argument parsing in Edalize and move it into FuseSoC. Arguments are currently parsed at two points in Edalize (`configure` and `run`) which will be handled in the following way:

`configure`: FuseSoC (or any other applicable frontend) will parse the arguments and write the parameter values to the `default` key for each parameter. This makes it no longer necessary to parse arguments on the Edalize side

`run` : FuseSoC (or any other applicable frontend) will parse the arguments and send a dict of key/value pairs to Edalize. Edalize will update its (run-time) parameter values accordingly (This also gives edalize the possibility to check/warn if any build-time arguments were changed without a rebuild)

Here's the full plan
1. Get rid of the `cli_parser` special case for verilator by introducing an extra_args/runtime_args argument in EDAM to allow passing arguments to verilated models that aren't parsed by FuseSoC/Edalize and deprecate the `cli_parser` option. Update FuseSoC migration guide
2. Add argument parsing to FuseSoC to set parsed values in EDAM file. Send an empty argument list to Edalize.configure
3. Release a new FuseSoC version 1.10
4. Make Edalize changes
4.1 Add a setup.py blocker on FuseSoC<1.10
4.2 Drop argument parsing from configure
4.3 In Edalize.run, check if incoming arguments is a list or dict. If list, run parse_args. If dict, assume key/value pairs and update internal parameter values.
4.4 Update example in README.md
4.5 Release edalize 0.2 and communicate API break
5. Back in FuseSoC, depend on edalize >= 0.2 and send parsed key/value pairs to edalize.run

olofk:
FYI, I have already implemented 2 but not pushed it yet

olofk:
1 is pushed to Edalize now. Added a warning instead of updating the migration guide

olofk:
Pushed 2 now #yolo #carpediem

olofk:
3 done. Time to work on the Edalize side

DHWinterWolf:
[...]

PR #380, opened 2020-03-22
Don't re-parse arguemnts in run

DHWinterWolf:
At present, with FuseSoC and Edalize at master, any backend arguments that take strings as parameters get truncated to their first character. I think the re-parse is unnecessary, as everything is done once at the beginning, and the bug is shown only in a re-parse during the run step.

I think this is a change that was missing in https://github.com/olofk/fusesoc/issues/363 - I've tried a few things locally with no problems, but I don't tend to use arguments relevant to the run step, so I'd appreciate others' opinions.

### DIFF
--- a/fusesoc/main.py
+++ b/fusesoc/main.py
@@ -478,7 +478,7 @@ def run_backend(
 
     if do_run:
         try:
-            backend.run(backendargs)
+            backend.run([])
         except RuntimeError as e:
             logger.error("Failed to run {} : {}".format(str(core.name), str(e)))
             exit(1)


olofk:
Sorry for the delay. This fix won't work unfortunately since it breaks overriding arguments during runtime. I had a look at a proper fix and it turns out to be more complicated than I first thought. A question about your original problem. Have you made sure to use FuseSoC 1.10 together with edalize 0.2.0 ? I can't really see the issue on my side

DHWinterWolf:
You raised an interesting point about having the right versions. So, I've just installed from scratch, definitely with FuseSoC 1.10 and Edalize 0.2.0. I do still see the problem.

To (hopefully) help, I've created a very simple test case to illustrate the problem. IT's at https://github.com/DHWinterWolf/fusesoc/tree/demo-pnr-problem/reparse-test-case - If you check that out and change to that directory, then run `fusesoc --cores-root . run demo`, output looks like this:

```
INFO: Preparing ::demo:0.0.0
INFO: Setting up project

INFO: Building
quartus_sh  -t demo_0_0_0.tcl
Info: *******************************************************************
[...]
    Info: Total CPU time (on all processors): 00:00:00
quartus_map  demo_0_0_0
Info: *******************************************************************
[...]
Info: Quartus Prime Analysis & Synthesis was successful. 0 errors, 1 warning
    Info: Peak virtual memory: 993 megabytes
    Info: Processing ended: Fri Apr  3 09:39:20 2020
    Info: Elapsed time: 00:00:09
    Info: Total CPU time (on all processors): 00:00:17
INFO: Running
```

So it's paid attention to the pnr: none in the demo.core file. However, if you run `fusesoc --cores-root . run demo --pnr none` (or indeed any other argument after `--pnr`, you get:

```
INFO: Preparing ::demo:0.0.0
INFO: Setting up project

INFO: Building
quartus_sh  -t demo_0_0_0.tcl
Info: *******************************************************************
[...]
    Info: Total CPU time (on all processors): 00:00:00
quartus_map  demo_0_0_0
Info: *******************************************************************
[...]
Info: Quartus Prime Analysis & Synthesis was successful. 0 errors, 1 warning
    Info: Peak virtual memory: 993 megabytes
    Info: Processing ended: Fri Apr  3 09:42:11 2020
    Info: Elapsed time: 00:00:09
    Info: Total CPU time (on all processors): 00:00:17
INFO: Running
Error (213013): Programming hardware cable not detected
ERROR: Failed to run ::demo:0.0.0 : 'quartus_pgm' exited with an error code
```

So it has not run P&R, but the parsing of that argument in the run stage (which should skip programming) has failed. Adding debug code shows that it interprets pnr as "n" in this example. Does that make sense?

olofk:
Many thanks for the test case and detailed analysis. I finally found time to take a look and I'm starting to understand what's going on. I think I have found the correct fix for this, but it's all a bit too messy so I don't want to push it yet. It would be awesome if you could test the `pnrfix`branch and see if that behaves as you expect

DHWinterWolf:
> Many thanks for the test case and detailed analysis. I finally found time to take a look and I'm starting to understand what's going on. I think I have found the correct fix for this, but it's all a bit too messy so I don't want to push it yet. It would be awesome if you could test the `pnrfix`branch and see if that behaves as you expect

Thanks very much for looking at that, Olof. The fix looks great, nice and simple and makes a lot of sense! I can confirm that I've tested it and it seems to do the job perfectly, thanks very much. Let me know if I can help in any other way.

olofk:
Perfect! Thanks for the quick testing. Pushed to master now.

PR #388, opened 2020-04-02
Don't use a class variable for _libraries in LibraryManager

rswarbrick:
This should have no effect when running the fusesoc program, but the
class variable can cause problems with test code (which doesn't
restart Python between different tests) because different the state
hangs around.


olofk:
I think the most likely answer is that piece of code was written before [this](https://twitter.com/OlofKindgren/status/1200177593875611648?s=19)

imphil:
:)

PR #389, opened 2020-04-02
Fix reading of EDAM file without configure

imphil:
In 57e50f282100489d8635068e13467fbe324306d0 the YAML reading
functionality got factored out into fusesoc.utils, but the import wasn't
done properly. Fix that.

Reported by Satnam Singh.

Fixes lowrisc/opentitan#1874


olofk:
Yep. I just did the exact same thing locally earlier today. Please commit

PR #390, opened 2020-04-06
Refactor Edalizer class

imphil:
Refactor this class to split out functionality into smaller sub-functions
to make it easier to extend. In the process, remove some Python 2
functionality we don't need any more. The call to the depenceny resolver
moved into the Edalize class as well to provide a slightly cleaner
interface: a CoreManager instance, and the top-level core.

Updated the tests to fulfill the new requirement to pass a CoreManager 
instance to Edalizer. Move some test files into separate directories to 
keep core libraries separate and avoid spilling between tests.

Thanks to Rupert for working on the tests.

Co-authored-by: Rupert Swarbrick <rswarbrick@lowrisc.org>


imphil:
This a preparation PR for subsequent ones that actually will change behavior. To make them easier to review let's do this one first. No change in behavior expected.

olofk:
I trust your judgement on this. Please be aware however that I'm also messing around inside the edalizer module to fix some bugs that were caused by the argument parsing changes. Most notably, pretty much all of the functions related to argument handling will be moved out from the edalizer module. Please let me know if that clashes with any of your upcoming changes

PR #391, opened 2020-04-07
Resolve dependencies for generator-created cores

### LINKED ISSUE #365: Generators: Resolve dependencies in generated cores
Generators can create new cores. Currently, the source files listed in these cores are inserted into the file list, but the dependencies (as listed in the `depend` statements) ignored. This issue is to change that.

The main challenge here is runtime: solving the whole dependency tree takes multiple 100 ms. So we don't want to do that after each generator run, or fusesoc runs will become quite slow with multiple generators present. Instead, I'm proposing to do one solving step after all generators have run. Upside: speed. Downside: later generators (e.g. generator n + 1) cannot see the files added by generator n. But that's probably acceptable, or even preferred (since it gives each generator the same view of the tree, no matter when it runs).

I've implemented this in https://github.com/lowRISC/fusesoc/commits/generators and am playing around with it a bit more before doing a PR for it. Most of the implementation is a cleanup of the `Edalizer` class to make it more modular.

imphil:
Currently, generated cores (i.e. core files written by a `generator`)
are inserted into the list of cores, but their dependencies are ignored.
After this commit, cores are inserted into the list of available cores,
and their dependencies are evaluated as with normal cores.

The implementation is actually simple:

- Generators produce files, including .core files, in a special
  directory, unique per generator run.
- This directory is added as library to the core database.
- All cores in this new library (i.e. all .core files) are added as
  dependency to the current project.

With that in place, the normal dependency resolution process takes over.

This new behavior is actually easier to understand for users, as
generator-produced are no longer a special case, but handled exactly as
other cores.

Fixes #365


olofk:
This looks surprisingly clean, which is encouraging. Does it still work with the `pos` argument for generators? I think I understand how the new resolving works, but 'm not 100% sure so would be good to confirm

imphil:
> Does it still work with the pos argument for generators?

Good question, I'm not sure any more. I'll have a look and make sure to add a test if not. (I also think I remember `pos` wasn't terribly well documented). 

I'll get back to you on the coming days.

fusesoc/edalizer.py
```diff
@@ -378,7 +402,7 @@ def generate(self, cache_root):
             cache_root (str): The directory where to store the generated cores
 
         Returns:
-            list: Cores created by the generator
+            Libary: A Library with the generated files
```
rswarbrick, line 405: Nit: Typo ("Library")

benreynwar:
What's the status with this?  I'd like to fix an issue with out-of-order files when using generators (#409) but was waiting to see what happened with this pull request.

imphil:
It's my turn to understand fully one side effect (related to ordering). To do so I got slightly side tracked, but I hope to finish this one off rather soon.

imphil:
Here we go. I updated this PR on top of @benreynwar's test improvements, and did a bit more digging around the position attribute of generators. There are three things which need addressing:
* position is untested so far.
* position will most likely be unsupported/not needed any more with this change. I need to properly test and document that.
* Somehow CI is failing, but not for me locally. Could be a Python version issue, I'll do some digging.

imphil:
OK, so `position` will not be supported any more after this change merges, causing an incompatible change to the current behavior. I'm not sure how many people actually use position with generators and what the fallout will be, but to be on the safe side, I'm not proposing this change for the fusesoc 1.* series, but for the next major release (2.*) to match semver expectations.

olofk:
FTR, I'm using position in several cases. One example is SweRV where a script is generating the CPU configuration, which must be added before the core. Note, this uses `first` but would technically be fine to use any position before SweRV. I have been meaning to add before/after as well. This would take care of some additional use cases but not the swerv case as the generator is likely called from a top-level core (like swervolf) that depends on swerv, so `before` would just mean before swervolf (but after swerv). I believe Ben is using generators to create VHDL packages which must be compiled before they are used. Again, if the ttptttg is defined in the toplevel core but the generator itself resides in a depency, we have the same issue.

Generally, I really think we should avoid allowing generators to inject cores in the dependency tree. It just seems to be too many corner cases to handle, especially with the curent dependency manager. It would be good to get an idea of the actual use cases. I suspect that some of them can be handled once the use flags mechanism is somewhat improved

imphil:
Yeah. Since this PR I have done a fair amount of thinking on the dependency manager and related things, and while I don't yet have a full solution to what we need in OpenTitan (which is motivating this PR), I think I'm getting closer. Let's keep this PR around for the time being, but don't try to get it merged just yet -- I'll come back once I've thought about this even more.

PR #395, opened 2020-04-30
Add support for Cadence Xcelium

tudortimi:


olofk:
I would prefer to not making any changes to CAPI1, since that has been deprecated for quite some time now. Did you make these changes for the sake of completeness, or do you have an actual use case for the capi1 changes?

tudortimi:
I took [https://github.com/olofk/fifo](fifo) and tried running it. I used `fusesoc sim`, because that was what I found in some example. I was too lazy to look for an example with `fusesoc run`.

tudortimi:
I don't need CAPI1.

olofk:
I suspected it was something like that and it's my bad for not porting my own cores to CAPI2. I'm dropping this PR but takes it as a reminder to fix my own cores

tudortimi:
Man, I spent like 10 minutes on this! That's time I'll never get back... (joke)

PR #400, opened 2020-05-07
Stringify args for calls to debug prints and subprocess

GCHQDeveloper992:
Whilst changing a core file I use from CAPI1 to CAPI2 I discovered some odd datatype behaviour that causes python to throw an exception in the utils.py file. 

Previously I used a CAPI1 core file with the name 'somecore-2019.11.core' representing version 2019.11 of 'somecore', using the year.month versioning scheme. With a new release, I thought I'd upgrade the core file to a CAPI2 one with a git provider section like:

```
provider:
  name: git
  repo: some_legal_git_url
  version: 2020.01
```

Unfortunately, this fails to check out because the version field is interpreted by the YAML safe_load function in capi2/core.py as a float. This gets passed through to the utils file whose
calls to join and check_call error as they expect strings and not floats.

This change makes sure that string representations of objects are used in the call to subprocess and join. I tried to work out if there was a better way of doing this earlier when the self.config object is created, but I'm afraid I got lost in the code. Does this seem a reasonable approach?

### DIFF
--- a/fusesoc/utils.py
+++ b/fusesoc/utils.py
@@ -27,7 +27,7 @@ class Launcher:
         logger.debug("    " + str(self))
         try:
             subprocess.check_call(
-                [self.cmd] + self.args, cwd=self.cwd, stdin=subprocess.PIPE
+                map(str, [self.cmd] + self.args), cwd=self.cwd, stdin=subprocess.PIPE
             ),
         except FileNotFoundError:
             raise RuntimeError(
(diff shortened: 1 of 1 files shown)


imphil:
The float parsing is part of the YAML spec; so the correct fix would be to do something like

```yaml
provider:
  name: git
  repo: some_legal_git_url
  version: "2020.01"
```

However, your looks like a good "safety net", so I'm good with taking this patch.

I know we don't have a test for this behavior yet, but can you add a doc comment to the `run` function explaining the behavior ("all arguments are converted to string")?

GCHQDeveloper992:
Thanks for the tip about quoting the version number @imphil , I'll update our core files to reflect that. I've added a quick doc string to run as suggested.

PR #402, opened 2020-05-22
Tidy CAPI2 tests

tudortimi:
This is a draft to show how we can make the CAPI2 tests more readable, by not having them rely on external files. This way, it's possible to figure out what the test is checking by simply reading the test body.

### DIFF
--- a/tests/test_capi2.py
+++ b/tests/test_capi2.py
@@ -38,9 +38,6 @@ filesets:
       - subdir/dummy.extra: {copyto : copied.file, file_type : user}
       - dummy.tcl: {copyto : subdir/another.file}
     file_type : tclSource
-  unusedfiles:
-    files:
-      - idontexist
   miscfiles:
     files:
       - vlogfile: {file_type : verilogSource, is_include_file : true}
@@ -76,8 +73,6 @@ targets:
     hooks:
       pre_build: [script_with_fileset]
     vpi : [vpi_with_fileset]
-  will_fail:
-    filesets : [unusedfiles]
   files_root_test:
     filesets: [fileset_in_files_root]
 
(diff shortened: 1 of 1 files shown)


tudortimi:
`test_capi2_export` and its core file are way too big. They could be split into multiple tests. I did an example of a split in d299ff0.

olofk:
You have a point. I am not very experience in making good test suites so I'm happy for any input on how to make these more usable.  I guess the drawback would be that it's harder to reuse files, but I'm not sure that really matters.

tudortimi:
https://medium.com/@kentbeck_7670/test-desiderata-94150638a4b3 states that tests should be readable.

I think this change enhances readability a lot. Of course, we lose some writability with this. It's possible to get to a middle ground, where we can reuse core descriptions using fixtures, to trade off some of the readability for writability (e.g. imagine creating strings/templates with the content of the core descriptions we want to reuse).

tudortimi:
Should I finish this change?

imphil:
I'm not quite convinced this goes into the right direction.

We should refactor our tests to not exercise the whole pipeline, but only a part of it: core file discovery, parsing, dependency resolution, etc. So for a discovery test, we need some actual core files in a file system. For parsing tests, we potentially need them, or can inline the code (as this PR suggests; I'm fine either way). For dependency resolution tests and other tests, we should just create `Core()` objects directly in the source code, and don't create them from CAPI2 files.

Unfortunately, doing so requires a bit of refactoring of the `Core()` class first, which currently is an interesting mix of parsing and dependency resolution code. I'm hoping to tackle that as soon as CAPI1 is gone.

@tudortimi are you OK with closing this PR for now, and I'll ping you when I've done the first part of refactoring and we can then potentially tackle the test cleanup together?

tudortimi:
@imphil Sure. It's anyway really out of date.

PR #405, opened 2020-06-11
Ignore .git subdirectories

Jbalkind:
We've talked about this before. For some of our old repos with big histories, fusesoc ends up spending a lot of time (~minutes) looking through .git subdirectories. I've had a tab open meaning to fix this but was finally pushed to do it.

Note that I made the exclude directories a set so we can add other things we should exclude.


olofk:
All good. Many thanks. We could probably add .svn and other stuff there as well. Maybe there's even a semi-standard list of these dirs. But .git is likely the most important so it's good we got that taken care of

PR #409, opened 2020-06-16
Fix regression in the file order of designs with generators.

benreynwar:
Recent refactoring of the Edalizer class introduced a bug where the files in the generated .eda are not in a dependency-consistent order.

This fixes this bug, by keeping track of how the generated cores are positioned in the ordered list of cores.

### DIFF
--- a/tests/capi2_cores/deptree/generated_child_a.core
+++ b/tests/capi2_cores/deptree/generated_child_a.core
@@ -1,0 +1,31 @@
+CAPI=2:
+# Copyright FuseSoC contributors
+# Licensed under the 2-Clause BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-2-Clause
+
+name: ::deptree-child-a
+
+filesets:
+  fs1:
+    files:
+      - "child-a2.sv"
+    file_type: systemVerilogSource
+
+    depend:
+      - ::deptree-child4
+
+generate:
+  generated-child-a-generate:
+    generator: generated-child-a-generator
+
+generators:
+  generated-child-a-generator:
+    interpreter: python3
+    command: generated_child_a.py
+
+targets:
+  default:
+    filesets:
+      - fs1
+    generate:
+      - generated-child-a-generate
--- a/tests/test_coremanager.py
+++ b/tests/test_coremanager.py
@@ -5,12 +5,15 @@
 import pytest
 
 
-def test_deptree():
+def test_deptree(tmp_path):
+    import os
     from fusesoc.coremanager import CoreManager
     from fusesoc.config import Config
     from fusesoc.librarymanager import Library
     from fusesoc.vlnv import Vlnv
-    import os
+    from fusesoc.edalizer import Edalizer
+
+    flags = {"tool": "icarus"}
 
     tests_dir = os.path.dirname(__file__)
     deptree_cores_dir = os.path.join(tests_dir, "capi2_cores", "deptree")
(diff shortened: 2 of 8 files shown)


olofk:
Thanks. Before applying it, could @imphil just take a quick look too, since I think you did the refactoring. IIRC this change was done in preparation of some other work. Just want to make sure that doesn't clash

imphil:
@benreynwar Do you think you can produce a test for that? We do have test_deptree in https://github.com/olofk/fusesoc/blob/master/tests/test_coremanager.py#L8, but it doesn't contain a generator. Would you mind extending that to cover this case as well?

benreynwar:
Finally taking a look at this.  It looks like currently the test is expecting the files in core child2 to be listed before child3.  But I would have thought (and the root.core file says) that we should expect the child3 files first.  Any objections to me going ahead and fixing this at the same time?

imphil:
I need to have a closer look tomorrow, but generally, the order in which dependencies are listed within a single `depend` block does not matter/guarantee any ordering. (Otherwise, the order in which cores appear in a `depend` file would itself create a dependency from item n to n+1 in the `depend` block.)

The only guarantees are: 
- Children come before parents.
- Files within a fileset are kept in order.

I think I remember that the test_coremanager test was rather fragile (or even partially wrong) and probably testing too much implementation details. Cleanups and fixes to that are very welcome, of course. Feel free to do them within this PR, but preferably within a separate commit.

benreynwar:
Yep, I was coming to the same conclusion looking at it some more.  I'll fix it up a bit.

benreynwar:
I think good guarantees are:

* Children before parents
* Files within filesets remain in order
* Order of filesets is unchanged.
* Files within generator remain in order
* Order of generators is unchanged.
* Files from generators come after files from filesets.

tests/test_coremanager.py
```diff
@@ -6,11 +6,16 @@
...
+    import tempfile
     from fusesoc.coremanager import CoreManager
     from fusesoc.config import Config
     from fusesoc.librarymanager import Library
     from fusesoc.vlnv import Vlnv
-    import os
+    from fusesoc.edalizer import Edalizer
+
+    flags = {"tool": "icarus"}
+    temp_dir = tempfile.mkdtemp(prefix="deptree_")
```
imphil, line 18: Please delete this directory after the test run.

tests/capi2_cores/deptree/generated_child_a.core
```diff
@@ -0,0 +1,31 @@
...
+    depend:
+      - ::deptree-child4
+
+generate:
+  generated-child-a-generate:
+    generator: generated-child-a-generator
+
+generators:
+  generated-child-a-generator:
+    interpreter: python
```
imphil, line 23:
```suggestion
    interpreter: python3
```

tests/test_coremanager.py
```diff
@@ -21,35 +26,99 @@ def test_deptree():
...
+        toplevel=root_core.name,
+        flags=flags,
+        cache_root=cache_root,
+        work_root=work_root,
+        core_manager=cm,
+        )
+    edalizer.run()
+    edalized_filenames = [os.path.basename(f['name']) for f in edalizer.edalize['files']]
+
+    assert expected_filenames == edalized_filenames
```
imphil, line 120:
```suggestion
    assert edalized_filenames == expected_filenames
```

to make pytest's diff view correct

benreynwar:
Ah.  I hadn't noticed #391 go by.  That's a nice improvement!

Currently the solver works out what core versions satisfy all the dependency constraints.  When the generated core dependencies are analyzed does the solver enforce the already chosen core versions as hard constraints?

i.e.
We have three modules
A - depends on B > 2.0 and generates C
C - depends on B == 2.0

We run solver on initial dependencies and get A and B version 3.0.
Then we run the generator and get the core for C.
But now the version of B required for C is incompatible with the version that we've already chosen.  So we fail due to incompatible cores.

Am I misunderstanding how things are working?

benreynwar:
Let's say I have a core that contains a mixture of generated and not-generated files, and has dependencies on other cores.

The elegant solution is for the main core file itself to just define a 'generate', and then for the generated core to define the dependencies and to list the generated and not-generated files in a fileset.  (Perhaps the main core should also specify the dependencies so that the first round of the solver also catches them?).

I'll just modify my generators so that they take parameters which specify the dependencies and the non-generated files.  Then they can generate a core file that accurately describes the core.

In light of this change in approach, I think it doesn't make sense to merge this branch until #391 is merged into master.  Then I'll modify this branch to be consistent.

imphil:
(For those who follow along this PR: This is more a discussion related to #391.)


>  When the generated core dependencies are analyzed does the solver enforce the already chosen core versions as hard constraints?

No, there's no "special magic" happening. Generated cores are added to the pool of possible solutions, and the solver starts all over. 

Regarding your example:
"B > 2.0 and generates C, C - depends on B == 2.0" -- that would be a circular dependency which you should avoid, independent of the question of versions. If B generates C, then B depends on C. That's the implicit dependency that #391 introduces. Don't have generated cores depend on the core that generate it.

> The elegant solution is for the main core file itself to just define a 'generate', and then for the generated core to define the dependencies and to list the generated and not-generated files in a fileset. 

Yes, that works well. 

Here's a slightly tricky example from OpenTitan: We need to generate a wrapper around primitives, like a RAM block. This wrapper has a generic interface (a SRAM interface, basically), but can be implemented in various ways for different technologies (FPGAs need a different implementation than a specific ASIC technology).  
So we have a fusesoc generator, which produces:
- A SystemVerilog file which instantiates different RAM submodules (e.g. ram_asic, ram_fpga, etc.)
- A core file, which depends on the actual implementations.

The full story is at https://github.com/lowRISC/opentitan/blob/master/hw/ip/prim/README.md, but as an example, here are a couple pointers to files:

- The core with the generator in it: https://github.com/lowRISC/opentitan/blob/master/hw/ip/prim/prim_ram_1p.core
- The generator core: https://github.com/lowRISC/opentitan/blob/master/hw/ip/prim/primgen.core
- The generator itself: https://github.com/lowRISC/opentitan/blob/master/hw/ip/prim/util/primgen.py

imphil:
OK, here we go. I updated this PR with a version that contains addresses my review comments and a couple tiny fixes, but is otherwise Ben's code. Let's get this in to document the current state of affairs regarding generators. 

I do have code ready for #391 to update this test with the slightly different behavior we get in this PR. This should make it much easier to see what is actually changing.

imphil:
Oh man. CI in this PR (and local testing) passed, but after rebasing this PR into master CI fails (https://github.com/olofk/fusesoc/runs/1417908199). Let me figure out what is going on ...

imphil:
Seems like I dropped @benreynwar's actual fix and only picked the test changes in my rebasing action. It is a mystery to me why CI didn't pick that up, but the most likely explanation is that the test ran actually not against the PR, but against master.

I added the changes back in https://github.com/olofk/fusesoc/pull/444. Sorry for the confusion (well, I mostly confused myself).

PR #410, opened 2020-06-16
The pipe to the stdin of the Launcher process is removed. The pipe is…

benreynwar:
… not used, and it gets in the way of debugging the generator using stdin (i.e. pdb breakpoint in generator).

I didn't see any reason for this pipe so I'm assuming it's use is historical.  I often do pdb debugging of the generators and make this modification locally so that I can communicate with the generators using stdin like normal.

### DIFF
--- a/fusesoc/utils.py
+++ b/fusesoc/utils.py
@@ -33,7 +33,7 @@ class Launcher:
         logger.debug("    " + str(self))
         try:
             subprocess.check_call(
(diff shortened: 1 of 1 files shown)


olofk:
I remember looking at that some time ago and couldn't figure out if it was used or not. I never properly understood these things so it's quite likely it was never needed and I just added it  out of confusion. Let's kill it and see if anyone complains

Issue #383, opened 2020-03-31
Support paramtype "generic" for backend Vivado


ThomasHornschuh:
Not sure of this is an edalize instead of fusesoc issue:
When using VHDL based cores with Vivado/xsim paramtype generic is not supported and lead to an error message:
`WARNING:root:Parameter 'TestFile' has unsupported type 'generic' for requested backend `

Workaround at the moment is using "vlogparam", but then I get a depreciation warning when using ghdl. 

BTW: In addition it would be good to support the real datatype for VHDL generics.

GCHQDeveloper560:
Lack of `xsim` support for generics is an Edalize issue. Generics have been supported by the Vivado synthesis/place-and-route backend (`vivado`) for some time, but you're correct that they are not supported for the Vivado simulation backend (`xsim`). I've hopefully correctly added support in [my xsim_generic branch](https://github.com/GCHQDeveloper560/edalize/tree/xsim_generic).

Are you able to test this and see if it works for you? Hopefully installing that branch is as simple as
```
pip install -U git+https://github.com/GCHQDeveloper560/edalize.git@xsim_generic
```

Support for a floating point data type is a (separate) FuseSoC issue.

olofk:
@GCHQDeveloper560, your xsim_generic branch looks fine to me from a quick glance. Happy to let it go in if it seems to work.

Handling floating point natively is probably a good idea. I have used strings to pass floats before (e.g. [here](https://github.com/olofk/stream_utils/blob/master/stream_utils.core#L41). That worked with plusargs in icarus but it probably breaks for backends and parameter types that wants to quote strings but not floating points.

If it's possible to pass them as strings for now, that would be great. Otherwise we need to look into that eventually

GCHQDeveloper560:
I took a look at the FuseSoC end and unless I missed something it's a pretty simple addition. See (my real_genparam branch)[https://github.com/GCHQDeveloper560/fusesoc/tree/real_genparam]. Rather than me dreaming up a bogus core/HDL as a test does this work for your case @ThomasHornschuh?

ThomasHornschuh:
I tried you edalize version. Unfortunately it screwed everything up. I got the traceback:
````
File "/usr/local/bin/fusesoc", line 11, in <module>
    load_entry_point('fusesoc', 'console_scripts', 'fusesoc')()
  File "/home/thomas/development/fusesoc/fusesoc/main.py", line 830, in main
    args.func(cm, args)
[...]

PR #411, opened 2020-06-16
Add support for real generics/parameters

GCHQDeveloper560:
#383

### DIFF
--- a/fusesoc/capi2/core.py
+++ b/fusesoc/capi2/core.py
@@ -369,6 +369,11 @@ class Core:
                     return default
                 else:
                     return int(default, 0)
+            elif datatype == "real":
+                if type(default) == float:
+                    return default
+                else:
+                    return float(default)
             else:
                 return str(default)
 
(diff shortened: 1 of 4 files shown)


olofk:
Hold on... can we really pull this off without _any_ changes in Edalize? Looking at the code it really seems like it would work. I'm just surprised because it was never my intention to make that possible :)

Anyway, if it works, let's go with it and just make a small update in the EDAM documentation to mention the existence of the real type. I have a slight preference for calling these float instead of real, but I don't care all that much.

So, if you could just confirm that there are no corresponding Edalize changes that I need to apply first, and I'll merge this

DHWinterWolf:
I've run a few tests on this branch, and can confirm it works fine with current Edalize at master, at least as far as ModelSim is concerned.

DHWinterWolf:
And for what it's worth, float would be fine by me if there's a preference for that name.

olofk:
Thanks. Let's push this and keep it real! :)

PR #412, opened 2020-06-16
Let os.makedirs handle existing directories.

m-kru:
In some cases os.path.exists may return False,
even if the path exists. In such scenarios running
FuseSoc results with FileExistsError. Letting
os.makedirs handle existing directories avoid such
buggy corner cases.


olofk:
All good. Thanks

PR #413, opened 2020-06-22
Catch error when backend tool is unknown

rswarbrick:
We were calling `get_edatool(tool)` twice: once for parsing backend
arguments and then again to do the actual work. The first call wasn't
surrounded by a try/catch for `ImportError`. This meant that if you
asked for a bogus tool (with a typo in your corefile), a
`ModuleNotFoundError` would make it to the console.

This patch surrounds the call to `get_edatool` with the try/catch block.
It also uses the `backend_class` that we found, rather than calling
`get_edatool` again.

### DIFF
--- a/fusesoc/main.py
+++ b/fusesoc/main.py
@@ -412,7 +412,11 @@ def run_backend(
     if not os.path.exists(eda_api_file):
         do_configure = True
 
-    backend_class = get_edatool(tool)
+    try:
+        backend_class = get_edatool(tool)
+    except ImportError:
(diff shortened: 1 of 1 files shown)


wallento:
Oh, yes, thanks! I stumbled over that just yesterday with a typo. There are a couple of similar semantics in core files and arguments that create strange erros by the way. I put those I recently faced in a list and can cross this one out now :)

imphil:
Wait, we're not making it as hard as possible for users who use the tool wrong?!!11

PR #415, opened 2020-06-24
Fix system wide installation command

dominiksalvet:
Hello fusesoc developers! :wave: 

This PR is a quick one, hence its branch name - tell me if it would be a problem. I was just going through the tutorial and reached an unformatted code block, so I fixed it. Feel free to modify my source branch! :rocket:


imphil:
Good catch, thanks!

Issue #337, opened 2020-01-10
Drop Python 2-support 


imphil:
According to the `pip install -U pypistats --user && pypistats python_minor fusesoc` we do have .55% of our users on Python 2.7.

```
$ pypistats python_minor fusesoc
| category | percent | downloads |
|----------|--------:|----------:|
| 3.5      |  75.12% |    37,041 |
| 3.7      |  22.32% |    11,005 |
| null     |   0.97% |       480 |
| 3.6      |   0.92% |       455 |
| 2.7      |   0.55% |       269 |
| 3.8      |   0.06% |        32 |
| 3.4      |   0.05% |        25 |
| 2.6      |   0.01% |         4 |
| Total    |         |    49,311 |

Date range: 2019-07-13 - 2020-01-08
```

How do people feel about removing Python 2 support? (Old releases are still available with Python 2 support, of course.)

wallento:
Yeah, I like that idea and am all for it. The only concern I would have is that the target industry is absurd when it comes to ancient tool versions. The stats are clear, so lets go for it.

imphil:
A bit of research shows:

Ubuntu 16.04: Python 2.7.12 and 3.5.2
Ubuntu 18.04: Python 2.7.15 and Python 3.6.7
RHEL/CentOS 6: Python 2.6.6 (no Python 3)
RHEL/CentOS 7: Python 2.7.5 and Python 3.6.8

So: RHEL6 is already not supported without a [SCL](https://www.softwarecollections.org/en/) (which are rather easy to use). RHEL7 has a recent-enough version of Python 3.

(Slightly more painful is the support for Python 3.5 in Ubuntu 16.04, but that will stay with us for a bit longer.)

olofk:
Thanks for this. It's specifically the RHEL/CentOS users I've been most worried about but it looks like supporting 2.7 won't really to any difference for them. Come to think of it, how many of those 269 py2 downloads are from Travis/Appveyor? :)

Why do you think it's painful with py3.5? From my perspective I think it's perfectly fine to have that as a minmum version but perhaps I'm missing something

I'm ready to drop py2

wallento:
IMO type annotations have been added after 3.5 which are really useful for a clear codebase

wallento:
https://docs.python.org/3/whatsnew/3.6.html

olofk:
Agree that it would be nice but doesn't outweigh losing 3.5 support at this point

Jbalkind:
[...]

PR #419, opened 2020-06-29
Add support for specifying library version

GCHQDeveloper028:
This commit adds a new option sync-version to the config file and
the library subcommand for the cli tool. This option is passed through
to the provider used for a library to specify a version of the library
to use.
This change also updates the git provider to use the sync-version to
checkout a commitish on a git-provided library. This libraries to be
fixed at certian points, or follow specific branches (such as nightly or
stable)

### DIFF
--- a/fusesoc/config.py
+++ b/fusesoc/config.py
@@ -115,18 +115,13 @@ class Config:
                 logger.warning(_s.format(str(e), name))
                 continue
 
-            try:
-                sync_uri = config.get(section, "sync-uri")
-            except configparser.NoOptionError:
-                # sync-uri is absent for local libraries
-                sync_uri = None
+            sync_uri = config.get(section, "sync-uri", None)
+            sync_version = config.get(section, "sync-version", None)
+            sync_type = config.get(section, "sync-type", None)
 
-            try:
-                sync_type = config.get(section, "sync-type")
-            except configparser.NoOptionError:
-                # sync-uri is absent for local libraries
-                sync_type = None
-            libraries.append(Library(name, location, sync_type, sync_uri, auto_sync))
+            libraries.append(
+                Library(name, location, sync_type, sync_uri, sync_version, auto_sync)
+            )
         # Get the environment variable for further cores
         env_cores_root = []
         if os.getenv("FUSESOC_CORES"):
@@ -165,6 +160,10 @@ class Config:
 
         if library.sync_type:
             config.set(section_name, "sync-uri", library.sync_uri)
+
+            if library.sync_version is not None:
+                config.set(section_name, "sync-version", library.sync_version)
+
             config.set(section_name, "sync-type", library.sync_type)
             _auto_sync = "true" if library.auto_sync else "false"
             config.set(section_name, "auto-sync", _auto_sync)
--- a/fusesoc/config.py
+++ b/fusesoc/config.py
@@ -115,9 +115,9 @@ class Config:
                 logger.warning(_s.format(str(e), name))
                 continue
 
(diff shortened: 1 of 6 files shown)


fusesoc/config.py
```diff
@@ -124,18 +124,13 @@ def __init__(self, path=None, file=None):
...
-                sync_uri = None
+            sync_uri = config.get(section, "sync-uri", None)
+            sync_version = config.get(section, "sync-version", None)
+            sync_type = config.get(section, "sync-type", None)
 
-            try:
-                sync_type = config.get(section, "sync-type")
-            except configparser.NoOptionError:
-                # sync-uri is absent for local libraries
-                sync_type = None
```
GCHQDeveloper028, line 137: I neatened these up while I was here, as it looks like python2 support was dropped (#337), lmk if I shouldn't have :)

fusesoc/main.py
```diff
@@ -702,6 +708,7 @@ def parse_args():
         "--sync-type",
         help="The provider type for the library. Defaults to 'git'.",
         choices=["git", "local"],
+        dest="sync-type",
```
GCHQDeveloper028, line 711: Added dest as this argument was being set as `sync_type` by the parser so I don't think it would be used

fusesoc/utils.py
```diff
@@ -29,7 +29,7 @@ def __init__(self, cmd, args=[], cwd=None):
     def run(self):
         """Runs the cmd with args after converting them all to strings via str
         """
-        logger.debug(self.cwd)
+        logger.debug(self.cwd or "./")
```
GCHQDeveloper028, line 32: Was printing `Debug: None` when cwd was not set, can remove from the PR if this isn't desirable

tests/test_libraries.py
```diff
@@ -131,6 +131,32 @@ def test_library_add(caplog):
...
+
+    vars(args)["sync-type"] = "git"
+    vars(args)["sync-uri"] = sync_uri
+    vars(args)["sync-version"] = "capi2"
+    args.location = None
+
+    expected = """[library.fusesoc-cores]
+location = fusesoc_libraries/fusesoc-cores
+sync-uri = https://github.com/fusesoc/fusesoc-cores
+sync-version = capi2
```
GCHQDeveloper028, line 145: This isn't super nice as the test will fail if this branch doesn't exist, wasn't sure of the best approach as the test clones the repository

olofk:
Hi,
Thank you for the contribution. It took far too long for me to get to this. I think it's good to have this functionality. At the same time I have been debating (with myself, so only half a brain) if it's worth the added complexity or if it can be easily solved outside of FuseSoC. But I think it's worth it after all. On top of the use cases you presented I'm thinking of cases where I have added some experimental feature to a branch and just want to pass someone a one-liner so they can test out the library for themselves.

Since some time has passed since you originally wrote this, it seems there are now some merge conflicts. We are also closing in on a new release and I would prefer to get that done before merging this. So once the 1.12 release is out, would you mind rebasing this to master and I'll push it right away. If you're unable to, I'll do it myself eventually

GCHQDeveloper028:
Hello!
No worries, once 1.12 is out I'll get on fixing the conflicts :)

GCHQDeveloper028:
@olofk - I've fixed the conflict and I believe this should be good to merge, let me know if there are any other changes needed :)

GCHQDeveloper028:
@olofk / @imphil - is this mergable? Happy to pickup any further changes as needed :)

olofk:

olofk:
For posterity, I can mention that I managed to salvage these patches and they are now committed

PR #429, opened 2020-07-27
Remove deprecated sim and build commands

imphil:
The sim and build subcommands have been deprecated for quite a while now,
and users have been warned about that. Let's remove them.

Users are encouraged to use "fusesoc run" instead, which is a full 
replacement.


olofk:
ok, let's do it! Fire at will. As a side note, I really want to reuse the `build` command some time in the future to use for FPGA builds without downloading it to a device. That would make a lot more sense. But let's remove this first and we can revisit a repurposed build command later

PR #431, opened 2020-07-27
Deprecate "fusesoc pgm"

imphil:
The same functionality can be obtained through "fusesoc run". We did 
deprecate the other special cases (fusesoc build and fusesoc sim) quite a
while ago, but apparently forgot to do the same for "fusesoc pgm". Do that
now, and remove it in the next release.


olofk:
Bring it on!

PR #439, opened 2020-11-11
FuseSoC is now BSD-licensed, update setup.py

imphil:
This bit of meta-information was missed when doing the license
migration, adjust that to show the right information at PyPi.


olofk:
Thanks! Never thought about that one. We should aim for a release soon. Probably want to do an Edalize release first though

PR #440, opened 2020-11-11
Use GitHub Actions for CI

imphil:
It's much easier to maintain than the current CI setup, and faster as well. To make things even easier, I've added tox support, so that tests can be executed simply by running `tox` in the source root directory.


imphil:
I don't think the GitHub Actions output will show up here before it is merged, let me know if you're OK with me going forward with that @olofk.

imphil:
OK, now the results show up and look promising!

imphil:
(AppVeyor is expected to fail, it's removed in this PR and needs to be manually disabled after merging.)

imphil:
OK, now the run won't show up until we have merged this PR, as it now uses the config from the main repository.

PR #442, opened 2020-11-16
[doc] Rewrite documentation

imphil:
Rewrite the FuseSoC documentation with a focus on the user guide. The 
rewrite adds a lot of new content, but isn't complete yet: the advanced 
topics of the build system are yet to be covered, as are the package 
manager functionalities.

Goals of this documentation effort:

- Think hard about target audiences, and provide the right content
(split) for them.
- Split build system and package manager functionality clearly.
- Try to give only one recommended approach if there are multiple
options in FuseSoC. The other approaches should be mentioned in the
reference guide, but not in the user guide.

For example: In the user guide
- core names are always specified in full VLNV,
- dependency constraints are only given in a single form, and
- CAPI1 is not mentioned.

Since the information that was previously available within the tutorial is
now in the user guide, the tutorial was removed.


imphil:
@benreynwar @rswarbrick If you have a bit of spare time (yeah, I know), I'd appreciate if you could have a look at this content as well. I did try to incorporate a fair amount of "best practices" into the user guide, but I'm sure I missed a lot of it as well.

doc/source/user/build_system/core_files.rst
```diff
@@ -0,0 +1,257 @@
...
+The core name, version, and description
+---------------------------------------
+
+Each core has a name, given in the ``name`` key.
+Core names can be freely chosen, but need to follow a common structure called *VLNV*.
+:term:`VLNV` stands the four parts of a core name, which are separated by colon (``:``): Vendor, Library, Name, and Version.
+
+.. todo::
+
+   Specify version number requirements, and what happens if no version number is given.
```
olofk, line 110: Maybe add a note that semver is highly recommended (might require that at some point in the distant future)

doc/source/user/build_system/core_files.rst
```diff
@@ -0,0 +1,257 @@
...
+* ``files``: An ordered list of source files.
+  The list of source files is ordered: the files will be passed to the tool in exactly the given order.
+  This is important, for example, in SystemVerilog, where packages need to be compiled before they can be used by subsequent source files.
+* ``file_type``: The default file type for all files in the ``files`` list.
+* ``depend``: Dependencies on other cores. More on dependencies later.
+
+Source files
+~~~~~~~~~~~~
+
+Source files are resolved relative to the location of the core file and must be stored in the same directory as the core file, or in a subdirectory of it.
```
olofk, line 150: Should we mention already here that remote cores work somewhat differently?

doc/source/user/build_system/core_files.rst
```diff
@@ -0,0 +1,257 @@
...
+
+Within a target
+~~~~~~~~~~~~~~~
+
+Within each target block multiple keys determine what the target does.
+The most common keys are:
+
+* ``filesets``: An ordered list of file sets (source files) included in the target.
+* ``description`` (optional): A description of the target.
+* ``toplevel`` (optional): The name of the design toplevel.
```
olofk, line 211: Maybe add a note that this can be a list, for designs which require multiple toplevels. Don't want it to be too prominent since it's somewhat rare (although many designs using Xilinx IP need this)

doc/source/user/build_system/dependencies.rst
```diff
@@ -0,0 +1,175 @@
...
+     - at least (greater than or equal to)
+     - ``>=fusesoc:examples:blinky:1.2``: version 1.2, or any newer version, e.g. 1.2, or 10.7
+   * - ``>``
+     - more (higher) than
+     - ``>fusesoc:examples:blinky:1.2``: any version after 1.2, e.g. 1.3, or 10.7
+   * - ``^``
+     - Caret requirement: any version less than the next major version (see below)
+     - ``^fusesoc:examples:blinky:1.2``: >=1.2.0 <1.3.0
+   * - ``~``
+     - Tilde requirement: allow updates to the current version (see below)
```
olofk, line 126: current _minor_ version ??

doc/source/user/cli.rst
```diff
@@ -0,0 +1,45 @@
...
+***************
+Running FuseSoC
+***************
+
+FuseSoC is a command-line tool; this section explains how to use it.
+The following content is aiming at users who already have a hardware design which uses FuseSoC.
+
+.. todo::
+
+   Think about giving users a demo/tutorial design to download and play with, which is also used in this section.
```
olofk, line 12: We could link to LED to Believe which already support a lot of boards if the users want to try it out, or add their own board otherwise. From LED to Believe there are also links to SERV as a next step

olofk:
Fantastic! Only had some minor comments and I'm happy to pull this in without them being fixed immediately. Well done!

doc/source/user/cli.rst
```diff
@@ -0,0 +1,45 @@
...
+      --build               Execute build stage
+      --run                 Execute run stage
+      --target TARGET       Override default target
+      --tool TOOL           Override default tool for target
+      --flag FLAG           Set custom use flags. Can be specified multiple times
+      --system-name SYSTEM_NAME
+                            Override default VLNV name for system
+
+
+Query
```
olofk, line 43: Is it supposed to look like this or is it a placeholder? Would be great to show users how to get a list of found cores quite early so they can verify that it works

doc/source/user/build_system/core_files.rst
imphil: I have ignored the package manager-like topics so far and haven't really thought about how to best phrase things to avoid confusion. I'd leave things as-is for now, and revisit when I do the package manager parts.

doc/source/user/build_system/core_files.rst
imphil: Good point. I thought about it and then didn't mention it, but I think we can easily mention it as advanced use. 

I've been thinking about always making it a list (to ease documentation and type hinting), but I'll do so separately once I have finished thinking (which can take a surprisingly long time) :smile:

doc/source/user/build_system/dependencies.rst
imphil: It's more tricky (confusing) than that, the text below describes it in more detail.

doc/source/user/cli.rst
imphil: This comment was left over from an earlier draft, I removed it for now while we figure out in more detail what a "best practices design" is.

doc/source/user/cli.rst
imphil: Oh, I just forgot to remove it. I planned to expand the `cli.rst` file more, but didn't get around to it (need to clean up some command line arguments first to actually be less confusing to document).

doc/source/user/build_system/core_files.rst
olofk: Please don't. It was a list originally and IIRC it was you who wanted to have it as a string. I think the current way is good as it makes the common case easy but still make the special case possible

doc/source/user/build_system/dependencies.rst
olofk: oh... I see.. Fair enough then.

doc/source/user/cli.rst
olofk: Yes. Documenting stuff has shown to be an excellent way to remove confusing code and options. Everything that's easier to remove than to document shall be gone :)

olofk:
Just squash and push when you're happy with it

doc/source/user/build_system/core_files.rst
imphil: Wait, I'm contradicting myself? Can't be!!!!!11

PR #444, opened 2020-11-18
CI: Run test in forked repo context

imphil:
Right now, we run pull request CI jobs in the context of the olofk/fusesoc
repository. We did this to be able to create Check runs, which show the
pytest results in the GitHub UI. There are multiple problems with that:

First, the runs actually didn't contain any data. That's probably fixable
if we found the right configuration regarding test result files and
artifacts.

Second, and more importantly, it seems that the "old" test code was run, 
i.e. pytest didn't pick up test changes done in the PR. That's hard to 
believe, but also not very useful, as PRs often modify code and tests at 
the same time. There might be some misconfiguration going on somewhere, but
I didn't debug it fully.

Finally, the test result reporting action we used is actually rather slow
(significantly slower than the test runs themselves), as it creates and
then runs a Docker container.

In this commit, we try a different approach and annotate the source code 
with test results.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,11 +13,11 @@ on:
 jobs:
   build-and-test:
     name: Build and Test (Python ${{ matrix.python-version }} on ${{ matrix.os }})
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     strategy:
       matrix:
         os:
-          - ubuntu-latest
+          - ubuntu-20.04
           - macos-latest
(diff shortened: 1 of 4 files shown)


imphil:
Nice! we now actually get test results inline in the code, e.g.

grafik

Currently, since we do matrix testing, each individual platform shows up and identical failures are not merged. That's not super beautiful, but probably better to show too much than too little. And not worth going in and developing a merging solution on our own, I'm sure someone will come up with something like that in the not too distant future.

PR #446, opened 2020-11-18
Test the position attribute for generate blocks

imphil:
The position attriute is meant to insert files generated by the core at
a specific location: after the chosen core, at the beginning of the file
list, or at the end. This behavior was not tested so far. Add a test for
it.

### DIFF
--- a/tests/capi2_cores/deptree/generated_child_a.core
+++ b/tests/capi2_cores/deptree/generated_child_a.core
@@ -17,6 +17,26 @@ filesets:
 generate:
   generated-child-a-generate:
     generator: generated-child-a-generator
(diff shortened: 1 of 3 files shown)


imphil:
This test is only to document the current behavior, which will change in #391. I want to ensure, however, that we have a very clear picture on what changes and ensure that we don't miss anything, so I'd like to get this test change in first.

imphil:
@olofk Can I go ahead and merge this? It's good to document the current behavior as I'm continuing to work in this area of fusesoc.

olofk:
Thanks! We're inching closer towards a new release

PR #447, opened 2020-11-19
Update index.rst

hackathlete:
Hello,
At least on my windows installation the initial fusesoc.conf file is located in the following Path: %HOMEPATH%\\.config\fusesoc\
Regards,
Markus

### DIFF
--- a/doc/source/user/package_manager/index.rst
+++ b/doc/source/user/package_manager/index.rst
@@ -17,12 +17,13 @@ file, ``fusesoc.conf``
 To find a configuration file, FuseSoC will first look for
 ``fusesoc.conf`` in the current directory, and if there is no file
 there, it will search next in ``$XDG_CONFIG_HOME/fusesoc`` (i.e.
-``~/.config/fusesoc`` on Linux and ``%LOCALAPPDATA%\fusesoc`` in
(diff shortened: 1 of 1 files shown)


imphil:
Thanks for this documentation update @hackathlete! 

Your fix is correct, we use `os.path.expanduser()` on all platforms, which is documented for Windows at https://docs.python.org/3/library/os.path.html#os.path.expanduser. 

I'm wondering, however, if the current behavior is actually what we want, or if we should put the config file into LOCALAPPDATA, where it probably belongs?

(In that case, we should still take this PR for now, but do a follow-up to migrate the config file to LOCALAPPDATA.)

hackathlete:
You're welcome @imphil and thanks for your explanation!

From a user perspective I agree, that %LOCALAPPDATA% is a better location as this would keep my %HOMEPATH% tidy. My %HOMEPATH% directory is cluttered with a lot folders from many different tools (for example .eclipse, .atom, .mozilla, ...). It seems, that using %HOMEPATH% is a common practice.

imphil:
Seems like many other applications originating from the Unix world have the same problem :-) I filed https://github.com/olofk/fusesoc/issues/448 to get this sorted properly, but for now, let's at least document what we have. Thank you!

Issue #361, opened 2020-03-02
Remove CAPI1 support


imphil:
The CAPI1 support is getting less and less testing, and the code paths are tricky to maintain. I'm opening this issue to see how we can deprecate and then at a later stage remove CAPI1 support from fusesoc.

Plan:
- [ ] Warn if CAPI1 cores are used. Do that for a bit of time. Documentation had a deprecation notice since Feb 2019, fusesoc will issue a deprecation warning with https://github.com/olofk/fusesoc/pull/453.
- [ ] Merge core converter: #188/ #231
- [ ] Remove CAPI1 support. Do a new major release (e.g. fusesoc 2.x). PR ready in #457

WDYT @olofk ?

olofk:
This has been on my mind for a long time. I agree with your item list, but would like to add two more items before them
-1. Deprecate `orpsoc-cores` for real by rewriting `fusesoc init` so it doesn't try to install it
0. Remove all CAPI1 cores from `fusesoc-cores`. If there is a version with CAPI2, just drop the older ones. Otherwise create a new CAPI2 version

ThomasHornschuh:
As a happy FuseSoC user with ~20 own cores I hope that you have some automated tool to convert CAPI1 to CAPI2 before removing CAPI1 support.  I'm aware that converting my cores is long overdue and assume that it will not take more than 1-2 days, but I'm always a bit reluctant to spend time for changes that does not give me an immediate added value in the moment I'm doing it.  In sum over all tools I'm using such things add up to more time I like to spend on it.

I assume that FuseSoC has an uniform in-memory representation of the core description. So I would find  some option like "write-corefile" which just write this representation into a CAPI2 the most convenient.

imphil:
There is unfortunately no "uninform in-memory representation", so a converter is significant work to be usable in a general case. If you want to contribute to that, there's a starting point in  #231.

ThomasHornschuh:
A short look into the PR shows that #231 seems to be a stand alone program. 
@olofk: What do you think is a better starting point, standalone program or add the functionality into FuseSoC?  
My understanding of FuseSoC code is that capi1 and capi2 be two Independent code paths. 
Eventually the capi1 path can become a "conversion only"  functionality. So a user can list and convert capi1 cores with it,  but not run them in a future version. 

Maybe to goal should be that the converter can handle "no-frills" simple cores and for all other be able to build a valid skeleton, containing dependencies, filesets, targets. 

[...]

PR #453, opened 2020-12-17
Issue deprecation warning for CAPI1

imphil:
The CAPI1 core description format has been deprecated for a while in our
documentation, but users of FuseSoC didn't, up to now, get a deprecation
warning when running `fusesoc`. Add such a warning to encourage users to
switch to CAPI2.

Commit deprecating CAPI1 in our docs: c03b095cf7d40e00d476ba5507c8e60a22286459 (Feb 2019)

Related to #361

### DIFF
--- a/fusesoc/core.py
+++ b/fusesoc/core.py
@@ -3,6 +3,7 @@
 # SPDX-License-Identifier: BSD-2-Clause
 
 import logging
+import warnings
 
 from fusesoc.capi1.core import Core as Capi1Core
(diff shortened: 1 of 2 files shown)


olofk:
Yes. Let's get this rolling. First thing however is to port all cores in fusesoc-cores so that we don't have a base library using deprecated functionality. That looks bad. There are about 15 left. Did a couple today. Help is appreciated

olofk:
Woohoo! Let's get it on!

Issue #264, opened 2019-05-08
Warn or error if fileset includes files above core file


imphil:
We should warn or error out if users have a core file which specifies a file path in a fileset with a relative path above the core file, e.g.

```yaml
filesets:
   file_rtl:
     files:
      - ../other_module/something.sv
```

Using fusesoc like this can work (if other_module for some reason part of the build as well, e.g. because of a dependency), but it doesn't work in the general case.

We should warn loudly or error out if this behavior is detected.

Sounds good @olofk? I'll open a PR then.

olofk:
Would it be enough to just check for `..` as the two first characters? Seems like it but I'm afraid I've missed something

imphil:
Checking for ".." as first file name component is sufficient (unless people use absolute paths. Maybe that should be disallowed as well)

imphil:
I opened https://github.com/olofk/fusesoc/pull/459 to issue a deprecation warning when paths outside of the core directory are found, with the intention to add the error in the next major release.

olofk:
We can close this now. The deprecation notice should be enough of a reminder for us. Reopen if you disagree

PR #459, opened 2020-12-18
Deprecate using files out of the core boundary

imphil:
The core file is expected to be at the root directory of an IP core, as
understood by FuseSoC; including files from files outside (above) this
directory was never really supported by FuseSoC, but we didn't error out
on it either. Add a deprecation warning with the intent to remove this
functionality in the next major release.

First part of #264

### DIFF
--- a/fusesoc/capi2/core.py
+++ b/fusesoc/capi2/core.py
@@ -6,6 +6,7 @@
 import logging
 import os
 import shutil
+import warnings
 import yaml
 
 from fusesoc import utils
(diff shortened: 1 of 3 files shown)


olofk:
Pinging @Jbalkind about this one as I have a vague memory of him (ab)using this feature

olofk:
Ah, it's just a warning for now. Cool

Jbalkind:
Hm I think if I was using this that it was because of some other deficiency that was maybe resolved by now. Can probably be solved by just adding more cores I expect.

olofk:
MORE CORES! MORE CORES! MORE CORES!

PR #460, opened 2020-12-18
Require Python 3.6

### LINKED ISSUE #454: Drop Python 3.6 support in FuseSoC 2.0
*In FuseSoC 2.0* we should drop support for Python 3.5, which has reached EOL in September 2020. 

One main benefit of doing so is significantly improved pathlib support (no more `str()` casts everywhere when using it), which will allow us to use more of its functionality when passing around filenames and paths in FuseSoC (which is currently typically done as string, and quite fragile at times).

The main downside is that we loose support for the "default Python" in Ubuntu 16.04 LTS, which will go out of standard support in April 2021 (which could match a FuseSoC 2.0 release quite well).

imphil:
Python 3.5 is EOL upstream; remove support for it from FuseSoC to enable
us to use newer features.

Fixes #454

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,6 @@ jobs:
           - macos-latest
           - windows-latest
         python-version:
-          - 3.5
           - 3.6
           - 3.7
           - 3.8
--- a/doc/source/user/installation.rst
+++ b/doc/source/user/installation.rst
@@ -12,7 +12,7 @@ System Requirements
 Before installing FuseSoC check your system requirements.
 
 - Operating System: Linux, Windows, macOS
-- Python 3.5 or newer.
+- Python 3.6 or newer.
   (The last version supporting Python 2.7 is FuseSoC 1.10.)
 - The Python packages ``setuptools`` and ``pip`` need to be installed for Python 3.
 
--- a/setup.py
+++ b/setup.py
@@ -56,6 +56,6 @@ setup(
         "pyyaml",
         "simplesat>=0.8.0",
     ],
-    # Supported Python versions: 3.5+
-    python_requires=">=3.5, <4",
+    # Supported Python versions: 3.6+
+    python_requires=">=3.6, <4",
 )
--- a/tox.ini
+++ b/tox.ini
@@ -50,7 +50,6 @@ commands =
 # (see the pipeline configuration) due to
 # https://github.com/ymyzk/tox-gh-actions/issues/44.
 python =
-    3.5: py3-ci
     3.6: py3-ci
     3.7: py3-ci
     3.8: py3-ci


imphil:
*Do not merge until `master` is FuseSoC 2.0*

imphil:
We can get this merged now with us working towards a 2.0 release.

Currently CI is broken on Python 3.5 due to setuptools_scm requiring now f strings from Python 3.6:

```
   File "/tmp/easy_install-ihbyp5t_/setuptools_scm-6.0.1/src/setuptools_scm/__init__.py", line 15, in <module>
  File "/tmp/easy_install-ihbyp5t_/setuptools_scm-6.0.1/src/setuptools_scm/version.py", line 42
    trace(f"tag '{tag}' parsed to {result}")
```

If we do not want to remove support for Python 3.5 we would need to fix the version dependency for setuptools_scm.

WDYT @olofk ?

olofk:
Tough one.

I'm not a big fan of raising requirements without good reasons. Even if Python 3.5 is EOL, e.g. Ubuntu 16.04 which ships with 3.5 is still supported to 2024. Not sure about the situation on Redhat.

OTOH we already decided to drop Python 3.5 support for FuseSoC 2.0 so it's bound to be done at some point anyway.

Didn't you fish out download statistics at some point? That could perhaps serve as guidance

imphil:
There are good reasons for Python 3.6+.
* The pathlib improved significantly, making it much easier to pass around Path objects without having `str()` casts everywhere. Operating on file paths is something we do extensively in FuseSoC.
* Ordered dictionaries are a massive help when operating on dicts that we want to dump out again, as we do a lot for CAPI parsing.
* Many smaller productivity improvements, e.g. f strings.
* Overall, the Python ecosystem is moving beyond Python 3.5, as the breakage caused by setuptools_scm shows. 

I argue that requiring Python 3.6 for the 2.0 FuseSoC release, expected in around half a year or so, will not prevent the vast majority of our users to update (if you want to continue using FuseSoC with Python 3.5, staying with FuseSoC 1.x is always an option).
* Almost no-one is using Python 3.5 with FuseSoC these days:
```
❯ pypistats python_minor fusesoc
| category | percent | downloads |
|----------|--------:|----------:|
| 3.8      |  81.10% |    13,889 |
| 3.6      |  15.25% |     2,612 |
| null     |   1.45% |       249 |
| 3.7      |   0.89% |       152 |
| 2.7      |   0.77% |       132 |
| 3.9      |   0.44% |        75 |
| 3.5      |   0.09% |        15 |
| 3.4      |   0.01% |         1 |
| Total    |         |    17,125 |

Date range: 2020-09-18 - 2021-03-17
```
* RHEL is typically not a problem as SCLs (software collections) are providing a convenient and commonly-used way to get newer versions of Python.
* For Ubuntu 16.04 will go out of standard support in April this year. Only the extended, paid-for support will go longer. If users really want to stay with this version of Ubuntu, PPAs and conda provide recent enough Python versions.

olofk:
Thanks. I'm convinced now. Wonder if those 15 downloads are from our CI :)
Picked and pushed

PR #461, opened 2020-12-18
Use f strings instead of .format()

imphil:
This gives us more readable and shorter string formatting.


imphil:
Requires a switch to Python 3.6+ first in #460.

imphil:
Rebased and ready for review.

olofk:
After carefully studying each changed string in detail I am happy to accept this

PR #462, opened 2020-12-28
Allow + characters in parsed expressions

sjalloq:
- added "+" to the list of special characters allowed in the parser.

### DIFF
--- a/fusesoc/capi2/exprs.py
+++ b/fusesoc/capi2/exprs.py
@@ -75,7 +75,7 @@ def _get_parser():
     if _PARSER is not None:
         return _PARSER
 
-    word = Word(alphanums + ":<>.[]_-,=~/^~")
+    word = Word(alphanums + ":<>.[]_-,=~/^~+")
     exprs = Forward()
 
     conditional = (
--- a/fusesoc/capi2/exprs.py
+++ b/fusesoc/capi2/exprs.py
@@ -12,7 +12,7 @@ FuseSoC core files allow strings matching the following pseudo-BNF:
   expr ::= word
          | conditional
 
-  word ::= [a-zA-Z0-9:<>.\[\]_-,=~/^]+    (one or more alphanum/special chars)
+  word ::= [a-zA-Z0-9:<>.\[\]_-,=~/^+]+    (one or more alphanum/special chars)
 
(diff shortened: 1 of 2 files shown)


imphil:
Thanks for your pull request. Can you describe what code triggered the need for this change?

sjalloq:
Hi Phil,

a certain incumbent with an area optimised CPU has directory names with a "+" in them.

Shareef.

imphil:
I see, thanks! I had a closer look and we need to think a bit more about the "word" definition in here, since theoretically any valid path name should be allowed (effectively any character). (And add tests for it.) But for now let's unblock your work and get this merged.

As a minimal change, would you mind updating the grammar definition in line 15 of the exprs.py file as well?

And there seems to be a mismatch between the grammar and the implementation with "~" appearing twice in the implementation. Can you align the two as well? (The additional backslash escaping in the grammar is intentional.)

CC @rswarbrick

sjalloq:
Thanks for the update.  I'll take a look at this sometime this week.

rswarbrick:
This looks good to me. As Philipp said, could you update the comment at the top? Also, please could you add a line to `test_exprs.py` with a plus in it? Maybe something like
```python
    check_parses_to("a+b", ["a+b"])
```
as the second line in `test_exprs`?

@imphil: I don't think we can allow arbitrary valid paths here. After all, they can contain things like spaces and newlines, which isn't going to play well with the rest of the syntax!

imphil:
> @imphil: I don't think we can allow arbitrary valid paths here. After all, they can contain things like spaces and newlines, which isn't going to play well with the rest of the syntax!
Merge state

Yeah, not sure if the current syntax is actually able to cover all edge cases. That's why I pushed it out into the "let's think about it some other day" area.

rswarbrick:
Yep, sounds sensible. I guess that if we want to handle arbitrary paths, we should allow an "escaped" syntax. Maybe allow a word to be a C-style string or something? But it's quite a bit more complexity, and I'm not sure it's really worth the effort.

sjalloq:
Think I've covered all the requests here.

imphil:
Thanks for your contribution! I combined the two commits into one and merged them.

PR #469, opened 2021-01-11
Add version information to documentation

imphil:
Add information about the version the documentation is built for to the
rendered documentation. This information is identical to the data shown
with `fusesoc --version` and ultimately derived from the tags in the Git
source tree by setuptools_scm.


imphil:
To pass on RTD https://github.com/olofk/fusesoc/pull/468 must be merged first.

PR #470, opened 2021-01-11
Use isort to sort all includes

imphil:
Isort enforces a single way to have includes sorted, which makes them slightly easier to read, but more importantly, removes the need to even think about it or do any formatting manually.


olofk:
Got a conflict. Perhaps easiest if you rebase this. Feel free to push as you wish

imphil:
:+1: rebased!

Issue #471, opened 2021-01-17
Can toplevel inherit name from the target name?


m-kru:
Almost always when I set `toplevel` within the target it has the same name as the target itself.
Example:
```yaml
  tb_start_after_reset:
    default_tool: ghdl
    toplevel: tb_start_after_reset
    filesets:
      - rtl
      - tb_start_after_reset
```

Is there any magic sequence (like `$`, `@`, `$1 `) to automatically assign the same name to the toplevel?
Example:
```yaml
  tb_start_after_reset:
    default_tool: ghdl
    toplevel: $
    filesets:
      - rtl
      - tb_start_after_reset
```

As you can see the same applies to one of the filesets.
The most compact version would be:
```yaml
  tb_start_after_reset:
    default_tool: ghdl
    toplevel: $
    filesets:
      - rtl
      - $
```

If it is not possible, are there any arguments against providing such functionality?

olofk:
I have been considering adding support for both user-defined (that can be set by use flags) and special variables (that are inferred by FuseSoC from the core description file). The use cases I have looked at would benefit from having name and version as special variables. I hadn't thought the target name, but that's a good idea. I don't think there's any yaml magic we can use for this so we have to do it ourselves. One thing that must be decided first is the syntax for variables.`$` doesn't work since we already support environment variables in core files. I have been considering %var% but I'm open to ideas. The important thing is that they don't clash with yaml syntax or other valid syntax in the core description files.

There's another approach to consider as well. I am normally very cautious when it comes to automatic things, but for toplevel, maybe we could consider setting it from the target name by default in case the user hasn't specified a toplevel themselves. Thoughts on that?

Also, related to this, there's a feature I introduced a while ago that might reduce clutter a bit if you have many targets like that. You can create a base testbench target with the common filesets, create an anchor and then use filesets_append, like this

```yaml
targets:
  tb_base: &base #This creates an anchor. The target only contain things that are common to all testbenches
    default_tool : ghdl
    filesets : [rtl]

  tb_start_after_reset:
    <<: *base #This brings in all the options from the tb_base target
    default_tool : modelsim  #Just an example if you want to override the default_tool from tb_base. Any options that is set here will override the same option in the tb_base target
    filesets_append : [tb_start_after_reset] #This appends the fileset the list of filesets that were inherited from the tb_base target
    toplevel : tb_start_after_reset
```

m-kru:
@olofk 
[...]

PR #473, opened 2021-01-21
Use target name as default toplevel name.

m-kru:
Related to https://github.com/olofk/fusesoc/issues/471

### DIFF
--- a/doc/source/user/build_system/core_files.rst
+++ b/doc/source/user/build_system/core_files.rst
@@ -206,6 +206,7 @@ The most common keys are:
 * ``filesets``: An ordered list of file sets (source files) included in the target.
 * ``description`` (optional): A description of the target.
 * ``toplevel`` (optional): The name of the design toplevel.
+  If the name of the toplevel is not specified, then the name of the target is implicitly used.
   (For advanced scenarios it is possible to specify a list of multiple toplevels instead of just a single one.)
 * ``default_tool`` (optional): The default tool to be used to build the target.
   The tool can also be set or overridden through a FuseSoC command-line argument.
--- a/fusesoc/capi2/core.py
+++ b/fusesoc/capi2/core.py
@@ -448,8 +448,12 @@ class Core:
             self._debug("Matched toplevel {}".format(toplevel))
             return " ".join(toplevel)
         else:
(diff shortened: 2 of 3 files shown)


imphil:
Thanks! Can you also mention this new behavior in the CAPI2 reference documentation at https://fusesoc.readthedocs.io/en/master/ref/capi2.html#target (it's generated from capi2/core.py).

m-kru:
@imphil done.

imphil (commented):
LGTM in terms of implementation. 

However, I'm not really convinced by this feature in the first place:
* It's a rather rare use case IMO, meaning we set a "default" which is rarely correct. In most cases, the target is named `synth` or `sim`, and the toplevel will have a very different name.
* This default will make it harder for users to see/debug cases where they failed to specify a toplevel, and now will get errors from their synthesis/simulation tool about e.g. the toplevel "sim" not found, something they never really specified. (Yes, there's an INFO Message, but EDA tool users have been trained over the decades to mostly ignore those.)

I leave the final call up to @olofk, however.

m-kru:
@imphil 

> It's a rather rare use case IMO, meaning we set a "default" which is rarely correct. In most cases, the target is named synth or sim, and the toplevel will have a very different name.

I think this is very user specific. In my most cases (90 %) the toplevel has the same name as the target.

> This default will make it harder for users to see/debug cases where they failed to specify a toplevel, and now will get errors from their synthesis/simulation tool about e.g. the toplevel "sim" not found, something they never really specified. (Yes, there's an INFO Message, but EDA tool users have been trained over the decades to mostly ignore those.)

I agree, I got the same feeling while implementing the change. A better approach would be to use some special sequence to refer to the target name like
```yaml
  toplevel: %target
```
or
```yaml
  toplevel: %target%
```

Yet another approach is to resign from this idea at all. Retyping the name for toplevel is not time consuming. Maybe it would be better to not break the orthogonality.

olofk:
Hmm... I didn't think about targets without toplevels. Maybe it's easiest to drop this idea and go start looking at adding variables to core files, so we can have e.g. `toplevel: %target%`. I would like to think this over for just a little bit. In any case we want to do this after releasing 1.12, so it won't go in now anyway

olofk:

PR #479, opened 2021-01-28
Target end users of stable FuseSoC in the README

imphil:
The README in the git repository is used for a variety of purposes, so we need to be careful about the content we put there.

1. It's shipped with the source code in the repository. 
2. It's displayed as "project description" on GitHub. 
3. It's used as package description in PyPi (at https://pypi.org/project/fusesoc/; frozen in time when the release was made) 
4. It's used as project description on LibreCores.

For all but the first use case people are actually expecting general information about FuseSoC, and not necessarily the latest information about
the current development version (or even information on how to develop it). To address this, this rewrite of the README

* assumes users want to install and use the latest stable version of FuseSoC,
* uses links which are valid for both the current stable release, as well as the upcoming stable release (which will see some restructuring in the docs),
* primarily targets users of FuseSoC, not developers,
* keeps the information in the README versioned and up-to-date by pointing to the versioned documentation instead.

As a editorial change, convert the README to Markdown, which is more common these days (sadly).


olofk:
Fuck yeah! Much better. Thanks

PR #480, opened 2021-01-28
Show actual version output for documented version

imphil:
When building the documentation we now execute `fusesoc --version` to
ensure that the documentation gives exactly the same output as the user
is expected to get, as long as they are reading the documentation
matching their installed version.

As a slight downside, we loose syntax highlighting for that block,
making it inconsistent with the other command snippets on the page. It's
fixable, but didn't seem worth it for now.


olofk:
Cool. As long as it doesn't break I'm fine with it

PR #483, opened 2021-01-30
Make strings with use flags explicit

imphil:
The CAPI2 documentation referred to the String type, which was
documented as always supporting use flags, i.e. all strings were
documented to be parsed. That's actually not the case. Supporting parsed
strings everywhere will make refactorings much harder, so this commit is
intending to keep the status quo instead of making use flags available
in more circumstances.

This commit introduces new data types, which also appear in the CAPI2
documentation, for strings which support use flags and which are parsed
as such when being used. The normal `String` type is now an unparsed
string, the new `StringWithUseFlags` is a parsed string.

Even though I hope to have caught all cases where a string is parsed in
the source code, I'm not 100% sure of that (mainly if the test suite has
coverage holes). To capture those cases in the field we throw an
exception in the hope that users report this behavior.

### DIFF
--- a/fusesoc/capi2/core.py
+++ b/fusesoc/capi2/core.py
@@ -28,14 +28,14 @@ class File:
         self.logical_name = ""
         if type(tree) is dict:
             for k, v in tree.items():
-                self.name = String(os.path.expandvars(k))
+                self.name = StringWithUseFlags(os.path.expandvars(k))
                 self.file_type = v.get("file_type", "")
                 self.is_include_file = v.get("is_include_file", False)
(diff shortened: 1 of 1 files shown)


imphil:
The rendered documentation is now much clearer on the support of use flags: https://fusesoc--483.org.readthedocs.build/en/483/ref/capi2.html

olofk:
Great! TBH, I have pretty much just added support for parsing use flags in different places when I have noticed it wasn't possible to do it, so it's highly likely there are more places that need to be changed. But in this case I think it's fine to wait until someone complains

PR #484, opened 2021-01-30
Ignore reformatting commit in git blame

imphil:
When using `git blame`, ignore commits which are only reformatting the
source code. This makes it easier to see who actually changed the line
functionally, instead of always referring to commits which reformatted
the whole source code.


olofk:
Didn't know you could do that. Neat

PR #486, opened 2021-01-30
Remove variable expansion from generate names

imphil:
The current code expands environment variables in
`targets.*.generate` blocks. This behavior is not documented and was
likely introduced in error in commit 4d02c089. Since we haven't released
this feature yet, let's remove it.

Example: before this patch, the following was possible, assuming
`TESTGEN_ENV_VAR` was set to `testgen`.

```
name : ::expandvars:0

generators:
  generator1:
    interpreter: python3
    command: testgen.py

targets:
  default:
    generate:
      - $TESTGEN_ENV_VAR
    toplevel: na

generate:
  testgen:
    generator: generator1
```

### DIFF
--- a/fusesoc/capi2/core.py
+++ b/fusesoc/capi2/core.py
@@ -80,9 +80,9 @@ class StringWithUseFlagsOrDict:
         if type(tree) is dict:
             for k, v in tree.items():
                 self.params = v
-                self.name = StringWithUseFlags(os.path.expandvars(k))
+                self.name = StringWithUseFlags(k)
         else:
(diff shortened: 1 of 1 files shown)


imphil:
This PR will conflict with #483, the second one to be merged will need to be rebased.

olofk:
Yeah, let's not support this. It will probably just add unncessesary headaches. Let's rebase and pull in after #483 is closed

imphil:
Rebased, and thanks to the disabled appveyor integration, all CI is now green as well.

PR #491, opened 2021-02-16
Better usability as module

wallento:
Currently, FuseSoC is only usable with its entrypoint. This adds the capability to use it as module by calling fusesoc.main.fusesoc(args). The parser can be instantiated separately to build the proper args object.


wallento:
The failed build doesn't really seem related..

olofk:
This looks fine to me. Thanks. Picked and pushed

PR #494, opened 2021-02-26
Remove the CAPI1 -> CAPI2 core converter

imphil:
The core converter is meant to be available only in FuseSoC 1.x, remote
it from master which is now targeting FuseSoC 2.x.


imphil:
Ready for review now as well.

Just to be clear, if we need to do a 1.13 release we can do so by using a stable branch, but I'd like to unblock master to get some CAPI2-only changes merged.

olofk:
I'm sceptic. The converter was only added two months ago and the new version is just a month old. Although most people are better off with the latest stable version, I'm pretty sure many install the development version too. I would like to have a far longer deprecation period than this. Can you explain what would be unblocked by this? If it's important enough I will reconsider

imphil:
We cannot remove CAPI1 before we remove the converter, and we cannot make progress on the internal representation of a core until we remove CAPI1. So none of the refactoring I have prepared for FuseSoC 2.0 cannot land on master until we go down that route.

If we want to make forward progress we need to get started. I don't see value in leaving the repository "dormant" for a while to cater for users who install the latest development release, but want it to be stable.

The deprecation period only matters for released code, there are no guarantees for code in a development branch. Everything we do in master today is unlikely to reach users in a stable release in less than a couple of months from now.

If at some point we have a desire to do a 1.13 release which includes some fixes users want before 2.0 can be released, we can always do so by creating a stable-1.x branch off the 1.12 release tag and cherry-pick commits into that.

olofk:
Ok, we can do this. But I think we need to write down deprecation policies somewhere. I would prefer to have a longer notice time in general. It's no fun for casual users to have their workflows broken without notice. At this point very few people would have a reason to update to 1.12 which means they haven't seen the deprecation warning

olofk:
And to add some more context to capi1 deprecation in particular. This should had been started a long time ago. The issue was that I never got around to convert the base library to capi2 which I wanted to do first. So ideally, the deprecation notice would had been added already a year or two ago.

I think this also highlights the need for a proper deprecation policy so we can easier identify blocking issues and have a better roadmap

imphil:
> So ideally, the deprecation notice would had been added already a year or two ago.

It was, just not as loud: the documentation mentions CAPI1 as deprecated for over two years (https://github.com/olofk/fusesoc/commit/c03b095cf7d40e00d476ba5507c8e60a22286459)

> It's no fun for casual users to have their workflows broken without notice.

I agree, but don't see how we are breaking users "without notice". 

* We did take great care to issue deprecation warnings for all things we know will be removed in the next major release.
* The 2.0 is a major release, with incompatible changes expected according to semver.
* If there are more things we want to deprecate before releasing a 2.0 release, we can do so by adding these warnings to a 1.13 release and make that available.

This gives users a very nice upgrade path: upgrade to the latest 1.x release, fix all warnings, and then upgrade to 2.0 and it should all continue to work. And as usual, if you don't want to upgrade, that's fine as well -- the old releases are still available.

I'm in general not a big fan of roadmaps -- they never work out. It is simply impossible to predict what a volunteer-driven open source project will be developing in half a year or even a year or two from now.  So instead of trying to make that half-work, just establish a development model that makes it easy for users to update, while at the same time not stalling the development.

An, as I find, excellent approach is what the Symfony project does and documents at https://symfony.com/doc/current/contributing/community/releases.html. This process is much more comprehensive than what we need for FuseSoC and can support, but it's showing rather well how users can be encouraged to upgrade while at the same time enabling experimentation and learning in the development.

Finally, look at the flip side of not moving forward: OpenTitan has been requiring a fork of FuseSoC for over a year now, because we require some features which cannot be implemented without breaking backwards-compat. If we don't want to end up with 10 forked versions of FuseSoC, we need a development model that allows those changes to be integrated into upstream in a reasonable timeframe.

olofk:
That Symfony page was an excellent read and very much sounds like what I would like to have for FuseSoC, although as you say a bit more lightweight perhaps.

I think you're right about the roadmap too. Current model works well enough. I'm convinced. Let's get it on

PR #500, opened 2021-04-06
Add errors key for targets to report error conditions

olofk:
This adds a new key to targets named 'errors' and a get_errors CAPI
function. Any item in the errors list that evaluates to a non-empty
string will be treated as an error condition and the string will be
printed out before exiting FuseSoC. Use-cases include erroring out for
unsupported combinations of flags, or disallowing the use of certain simulators

### DIFF
--- a/fusesoc/capi2/core.py
+++ b/fusesoc/capi2/core.py
@@ -335,6 +335,10 @@ class Core:
             depends += [Vlnv(d) for d in self._parse_list(flags, fs.depend)]
         return depends
 
+    def get_errors(self, flags):
+        target = self._get_target(flags)
+        return self._parse_list(flags, target.errors) if target else []
+
     def get_files(self, flags):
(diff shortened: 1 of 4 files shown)


imphil:
Hm, this approach feels a bit too rough to me and not explicit enough. 

Could you give an example use case to better illustrate the problem you're trying to solve? What we possibly want to specify is not "a list of errors", but a list of conditions under which a target is usable, and throw an error otherwise.

Another way to think about it: flags could be a requirement (as in something that the dependency solver takes care of) or a conflict.

olofk:
One common use for this would be to set `errors: ["!tool_verilator? (Testbench only supports Verilator)"]`
It could also be used to check for invalid combinations of use flags `errors: ["!spi? (boot_from_spi? (spi must be enabled to boot from SPI))"]`

PR #501, opened 2021-04-06
Disable Edalize verbose mode by default

olofk:
Disable tool output during the build phase by default. If build returns an error, the buffered stdout and stderr will be printed on the terminal. Running `fusesoc --verbose` reenables tool output


imphil (approved):
LGTM, thanks!

PR #502, opened 2021-04-06
Default use flags

olofk:


imphil:
Are flags inherited from the target? Can you add documentation for that to https://fusesoc.readthedocs.io/en/latest/user/build_system/dependencies.html#what-happens-if-a-dependency-is-specified?

(I just realized the docs aren't completely right, I filed https://github.com/olofk/fusesoc/issues/504)

And I'm not fully convinced about the tool being just another flag -- while we can make this simplification in the code, the tool is still conceptually something different than a flag, and has a very special meaning within FuseSoC.

olofk:
We can't do inheritance here. The flags must be evaluated first before we start the dependency resolution. In the future I envision we will have local use flags so that you can depend on e.g. `vendor:lib:some_cpu[fpu]` to require some_cpu to be built with an FPU but that's another story and won't conflict here.

Looking at it today however, there's another thing I want to change. I did a first iteration where the flags were stored as a dict instead but concluded a list was a better choice. I am considering a dict once again now. I chose a list to better support multiple values being set for the same option, but I see a way around that now. The list also makes it awkward to support the common case of selecting exactly one of multiple options and requires explicitly disabling the default value when you want to override it on the command-line.

Regarding the tool flag, I see your point, but I think we should try to treat it the same way as other flags. It just makes things easier.

I will cook up a new patch where the flags are implemented more like the parameters and probably also expand the CLI part a bit more so that it's possible to e.g. do `--flag option=value`

olofk:
I finally got around to update this. It still doesn't have improved CLI support but the target syntax is changed to a dict instead of a list. @imphil would be good to have your thoughts here. Perhaps also @mballance might be interested since we discussed this in https://github.com/olofk/fusesoc/issues/531 a while ago

olofk:
Merged

PR #506, opened 2021-05-07
Check if core name is empty string.

m-kru:


olofk:
Very good. Thanks for this. Picked and pushed

FYI, it's not a requirement but would be great to also have a test case added for this

PR #509, opened 2021-06-18
Setup and document the use of sphinx-autobuild

imphil:
sphinx-autobuild builds the documentation, serves it on a localhost web
server, automatically rebuilds it when changes are detected, and also
refreshes the web page automatically. With this setup documentation
writers can have the documentation open in their browser, change
documentation source code, and see the updated documentation as soon as
they save the source file, without any other action needed. Beautiful!
    
Add the tox setup for this, and document it as the preferred way of
working on documentation. The more manual way remains in place for now.

Additionally, this PR contains small two cleanup commits that I came across.


olofk:
Thanks. All good. I had no idea you could use tox for this :)

PR #510, opened 2021-06-18
Document use flags, and recommend quoted YAML strings

imphil:
This PR contains two commits which depend on each other, but are different topics:

- The first commit recommends to always quote strings used as value in CAPI2 files. This gives our users a single, always-right and easy to remember rule on what to do with strings. All other rules are more complex ("quote some strings, but not others") and easy to get wrong (as multiple support requests I've gotten over the recent months have shown).
- The second commit documents flags and their uses in FuseSoC, filling a gap in the documentation.

### DIFF
--- a/doc/source/user/build_system/core_files.rst
+++ b/doc/source/user/build_system/core_files.rst
@@ -9,6 +9,8 @@ Core files are written in :term:`YAML` syntax and follow the FuseSoC's own CAPI 
 Don't worry: using FuseSoC neither requires a full understanding of YAML, nor an up-front knowledge of CAPI.
 However, some key facts about YAML are important.
 
+.. _ug_build_system_core_files_yaml_intro:
+
 Things one should know about YAML
 ---------------------------------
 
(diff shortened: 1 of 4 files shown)


doc/source/user/build_system/core_files.rst
```diff
@@ -24,30 +26,44 @@ Things one should know about YAML
...
-     key2: value2
+     key1: 'value1'
+     key2: 'value2'
 
   In most cases, the longer (second) form is preferred, as it is easier to make changes while keeping the diff easy to read.
 
+* **We recommend always adding single quotes around strings used as value.**
+  In most cases, strings in YAML do not need to be surrounded by (single or double) quotation marks.
+  However, the exact rules when quoting is needed are not easily summarized without in-depth understanding of the YAML language syntax.
+  At minimum quotation marks are *required* when strings start with an exclamation mark.
```
imphil, line 56: Double quotes would work almost equally well for our use case, and might be slightly more "typical" for strings. No strong feelings there, I went with single quotes as they are slightly "safer".

doc/source/user/build_system/flags.rst
```diff
@@ -3,6 +3,182 @@
...
+Flags can be used in multiples places in :term:`core files <core file>` to take conditional actions.
+For example, dependencies and files can be included depending on a flag, different toplevels can be selected, and much more.
 
-   Document flags.
+Facts about flags:
+
+* Flags are boolean variables, they are either "set" (true) or "unset" (false).
+* Flags are global, i.e. they apply to the whole build.
+  All cores see the same flags under the same name.
+* A valid flag name starts with a letter, and consists only of letters from the English alphabet, numbers, and the underscore (``_``).
```
imphil, line 15: The naming rules are more strict than they need to be at the moment, but I thought it would be good to give us a bit of headroom for a different parser without restricting users too much -- these naming rules are already "common sense".

doc/source/user/build_system/core_files.rst
olofk: Me neither. I tend to use double quotes myself but that doesn't mean anyone else has to follow that

doc/source/user/build_system/flags.rst
olofk: Agree

doc/source/user/build_system/flags.rst
```diff
@@ -3,6 +3,182 @@
...
+Setting flags
+-------------
+
+Built-in flags are made available automatically.
+User-defined flags can be set on the command line and override built-in flags.
+
+Built-in flags
+~~~~~~~~~~~~~~
+
+When running ``fusesoc build`` some flags are automatically made available.
```
olofk, line 28: 'fusesoc run'?

doc/source/user/build_system/flags.rst
```diff
@@ -3,6 +3,182 @@
...
+  fusesoc build --flag "my_flag" --flag "my_other_flag" fusesoc:examples:blinky:1.0.0
+
+  # Unset the flag "my_flag"
+  fusesoc build --flag "-my_flag" fusesoc:examples:blinky:1.0.0
+
+.. note::
+
+  Order matters!
+  The FuseSoC command line is "context sensitive."
+  Place the ``--flags`` argument after the ``build`` command, but before the name of the core you are building.
```
olofk, line 67: run?

doc/source/user/build_system/core_files.rst
imphil: Yeah, my muscle memory also clearly prefers double quotes. I thought about it again and the benefits of using single quotes aren't really worth using the "less common" quotes for strings. I switched to double quotes instead.

imphil:
Thanks for your review @olofk! No idea how I confused myself into thinking that "fusesoc build" existed :-)

I also switched to recommending (and using) double quotes now, since we both seem to prefer it, and honestly the benefits don't really outweigh the more common quotation marks.

olofk:
Cool. No more complaints, so pushing now. Many thanks for this

PR #511, opened 2021-06-18
Remove Sphinx from dev-requirements.txt, add tox

imphil:
The recommended way of building the documentation locally is through
tox. Avoid installing Sphinx for "manual" development environments, tox
already takes care of that.

Add tox as development dependency (it was missing before), and remove
the manual installation of Sphinx.

For now we keep pytest as separate dependency, even though that's not
needed when running tests through tox.


olofk:
You know this stuff way better than I do so I'll just pull it in if you think it makes sense

PR #512, opened 2021-07-15
Added script to generate .f file listings from core files

TParry-SystematIC:
This script adds the ability to auto-generate .f style file listings from the a core file with the same syntax used for fusesoc.

Details on implementation specifics could be answered by the scripts original developer, @sjalloq.

### DIFF
--- a/extras/core2vc.py
+++ b/extras/core2vc.py
@@ -1,0 +1,92 @@
+#!/usr/bin/env python3
+
+# -----------------------------------------------------------------------------
+# This script uses the FuseSoc dependency builder to generate a list of files
+# used for a particular target in a core file.  This can be used to auto-
+# generate EDA style VC files without having to keep two sets of source file
+# lists up to date.
+# -----------------------------------------------------------------------------
+
+
+
+import os, argparse, subprocess, logging
+from pathlib import Path
+from coregen import CoreDeps
+
+logger = logging.getLogger(__name__)
+
+header = """
+// ----------------------------------------------------------------------
+//            This file is auto-generated using core2vc.py.
+//
+//                  -=  D O   N O T   E D I T  =-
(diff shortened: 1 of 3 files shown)


Jbalkind:
Does this run any generators that are part of the core file? Can you specify targets etc? I'm wondering if this might make more sense to be a kind of null backend for edalize?

sjalloq:
@Jbalkind, I need to look at how I wrote this again but I think your idea of an edalize backend might be more suitable.  I'll try and take a look next week.

imphil:
There has been some discussion about the question how to generate a `.f` file from FuseSoC runs in the past, and I think the outcome was the only working approach is using the EDAM file that FuseSoC generates (which is part of the FuseSoC API) and converting that into a .f file that suits the tool you're using. That should be easy to achieve in a simple Python script.

In the most trivial case, the script can look like this:

```python3
import yaml
import sys


def main():
    eda_yaml_file = sys.argv[1]
    with open(eda_yaml_file) as eda_yaml_f:
        eda_yaml = yaml.safe_load(eda_yaml_f)

    for f in eda_yaml["files"]:
        if f["file_type"] in ["systemVerilogSource", "verilogSource"]:
            print(f["name"])


if __name__ == "__main__":
    main()
```

Two reasons:
- `.f` files are not really standardized on "edge cases"; some tools (e.g. VCS) allow SystemVerilog parameters and defines in it, while others expect a plain list of source files.
- The tool is a global flag in FuseSoC and its taken into account when doing dependency resolution (c.f. https://fusesoc.readthedocs.io/en/latest/user/build_system/flags.html) and other things. Currently, a tool is equal to a backend. So if you'd create a new ".f file backend", you effectively create a new tool. It's unlikely that you want that.

olofk:

PR #516, opened 2021-08-03
Add Integer class for edalize

### LINKED ISSUE #515: Unknown item 'Integer'. Parse Error
Hi,

I am using FuseSoC with Vivado.  When I define a parameter with `Integer` type like `jtag_freq` or `jobs`, FuseSoC throws this error. Other `String` type settings like `part` works fine.  Is it a problem related to CAPI2 parser, `fusesoc/capi2/core.py` ? It looks like parser doesn't know `Integer` type settings. FuseSoC 1.12.0 is installed with edalize 0.2.5. Problem exists both on Windows and Linux.

A question: Is it possible to override `jobs` parameter of Vivado (or any parameter in general) from command line when calling FuseSoC. I want to set this parameter on Linux with `$nproc` while calling the FuseSoC like `make -j$(nproc)`.

Thanks

alperyazar:
This commit add basic Integer class to support Integer type parameters for back ends defined in edalize. For example: jobs parameter of Vivado. Please see Issue #515 https://github.com/olofk/fusesoc/issues/515

Fixes #515

### DIFF
--- a/fusesoc/capi2/core.py
+++ b/fusesoc/capi2/core.py
@@ -65,6 +65,8 @@ class String(str):
             "https://github.com/olofk/fusesoc/issues/new."
         )
 
+class Integer(int):
+    pass
 
 class StringWithUseFlagsOrList:
     def __new__(cls, *args, **kwargs):


alperyazar:
This PR solves my problem in #515. But the genreated `*_eda.yml`  may have some problem. Integer parameters exported as strings, for example in generated `*_eda.yml`:

```yaml
jobs: '8'
```

what I expect is

```yaml
jobs: 8
```

For the current version (0.2.5)  of edalize and Jinja engine this doesn't cause a problem. I would appreciate any help to fix this export mismatch.

The only tool with Integer parameter in edalize is Vivado and those parameters are `jobs` and `jtag_freq`.  `jobs` added with [this PR](https://github.com/olofk/edalize/pull/222) and `jtag_freq` added with [this commit](https://github.com/olofk/edalize/commit/5df35f85c2b4ae2da8f3930e8c09e6d6a5ae22fd). But it looks like FuseSoC have a problem with Integer, so are these parameters working before or am I the only one having problem with Integer type parameters. I would like to CC @m-kru and @benreynwar 

We may also consider adding a test to CI to cover this situation, a regression test.

Thank you.

olofk:
I hit this problem a while ago as well and I think I managed to fix it properly so that it turned up as integers in the EDAM file. Unfortunately I can't find the changes that I made and apparently never checked them in somewhere. Anyway, this fix isn't complete as @alperyazar has noted but I'm pulling it in anyway since it's at least better than before. Would be great to get these parameters to turn up as proper integers in the EDAM in a follow-up patch

And yes, test cases for these things would be good to have. Testing has fallen a bit behind lately :/

PR #524, opened 2021-09-30
Support virtual packages in CAPI2

### LINKED ISSUE #235: Choose between implementations of an interface
This is essentially a question/starting point for discussions about a use case which probably has happened before, and I'd like to get @olofk's thoughts on the topic. (I'm very sorry for the long text, and it's probably restating much stuff that is mostly clear, but better be on the safe side ...)

**Scenario**
(This is rather specific, you could do it with similar constructs like defines as well.)
- I have a SystemVerilog package which defines common parameters for a bus, like data and address width. Something like that:

  ```sv
  // modules/vendor/bus/rtl/bus_conf_pkg_default.sv
  package bus_conf_pkg;
    localparam BUS_AW = 32;
    localparam BUS_DW = 32;
  endpackage;
   ```

- Code in my bus module uses the `bus_conf_pkg` package as a source of global configuration, e.g.

  ```sv
  // modules/vendor/bus/rtl/bus_top.sv
  module bus_top(
    input [bus_conf_pkg::BUS_AW-1:0] addr
  );
  // ...
  endmodule
   ```
- I want to create a reusable module with a FuseSoC description. This means here: Give the user of my module the chance to configure the bus project wide by providing their own implementation of `bus_conf_pkg`. In practice this means the user should be required to include a FuseSoC core which provides the `bus_conf_pkg` package, but I don't want to force the user to take a specific one.

**Current FuseSoC solution**
- Create a core file which provides the `bus_conf_pkg` package.
  ```yaml
  CAPI=2:
  # modules/vendor/bus/config.core
  name: vendor:bus:config
  description: "The bus configuration"

  filesets:
    files_rtl:
      files:
        - rtl/bus_conf_pkg_default.sv
      file_type: systemVerilogSource
  ```

- Create another core file for the implementation which depends on a core providing the `bus_conf_pkg` package.

  ```yaml
  CAPI=2:
  # modules/vendor/bus/implementation.core
  name: vendor:bus:implementation
  description: "The bus implementation"

  filesets:
    files_rtl:
      files:
        - rtl/bus_top.sv
      file_type: systemVerilogSource

      depend:
        - vendor:bus:config
  ```

- I ship the module with both core files, and an implementation using the core adds a `depend: [ vendor:bus:implementation ]` to its core file and it can happily use the bus with the default values given in `bus_conf_pkg_default.sv`.

The challenge starts when one wants to override the default configuration, i.e. provide an own implementation of `bus_conf_pkg`. I see the following options currently:

1. Override the `vendor:bus:config` core with a customized core implementation. Since cores are referenced by name, this means I can create another core file equal to the `modules/vendor/bus/rtl/config.core` file, let's call it `system/bus_config.core`. For this file to override the `modules/vendor/bus/rtl/config.core` file it must be placed in the CORES_ROOT search path (before or after, I always forget) the other file. 

    This approach works with fusesoc today, but it's error prone: if the search path is messed up, the design ends up synthesized just fine, but with the wrong configuration values (as `bus_conf_pkg_default.sv` is picked up instead of the customized copy). Messing up the search path is easy, since this piece of configuration depends on the user's build environment.

2. Do not declare the dependency on the `bus_config_pkg` in FuseSoC, i.e. omit the `depend: [ vendor:bus:config ]` stance from the `modules/vendor/bus/implementation.core` core file. Users then explicitly need to include any core file which provides a package named `bus_config_pkg`, otherwise they get compile errors from their synthesis tool.

    Again, that works today and it's fairly safe to do, however it circumvents the dependency resolution in FuseSoC and relies on the backend tool instead. That's not ideal either.

Since none of the solutions that I came up with are ideal, I thought about a better solution.

**A better solution**

Goals:
- All dependencies should be stated explicitly in the core files.
- Do not rely on the ordering of directories in the search path to determine which cores are used.

The idea essentially is to introduce a `provides` option in the core file, which specifies an "interface" a core implements. Other cores which require an implementation of this interface specify the interface name in their `depend` line, which looks exactly as before.

```yaml
CAPI=2:
name: vendor:bus:implementation
description: "The bus implementation"

filesets:
  files_rtl:
    files:
      - rtl/bus_top.sv
    file_type: systemVerilogSource

    depend:
      - vendor:bus:config
```

```yaml
CAPI=2:
name: vendor:bus:config_default
provides: vendor:bus:config # new syntax here
description: "The default bus configuration"

filesets:
  files_rtl:
    files:
      - rtl/bus_conf_pkg.sv
    file_type: systemVerilogSource
```

```yaml
CAPI=2:
name: othervendor:project_x:busconfig
provides: vendor:bus:config # new syntax here
description: "The bus configuration specific to project_x"

filesets:
  files_rtl:
    files:
      - rtl/bus_conf_pkg.sv
    file_type: systemVerilogSource
```

Now to use the bus, you need to explicitly choose which "provider" (implementation) of `vendor:bus:config_default` you want to include:

```yaml
CAPI=2:
name: othervendor:project_x:top
description: "The toplevel core file of project_x"

depend:
- vendor:bus:implementation
- othervendor:project_x:busconfig
```

Final notes:
- To keep things simple I've specified `provides` as a string. It could also be an list if a single core implements multiple interfaces. That's probably overkill, however.
- `provides` is used as a keyword in RPM, for fusesoc `provide` is more in line with `depend` (which I always mis-spell as `depends` ...)

So, after reading this whole essay @olofk, are you going to tell me that something like this is already implemented and waiting to be used? :-)

imphil:
Packages can now provide more than one package: they provide itself (as
described by the package `name`), and a list of further (virtual)
packages, described by the `virtual` key in the root of the CAPI2 file.

Example:
```
CAPI=2:
name: ::impl2:0
virtual:
  - ::someinterface:0
```

This core provides a core named `::impl2:0` and a core named
`someinterface:0`. Other cores can now depend either `::impl2:0` or
`::someinterface` and always get the `::impl2:0` core.

The keyword `virtual` was chosen over alternatives, such as `provide` to
be distinct from other keywords we use in CAPI2 currently, especially to
be distinct from the concept of providers.

Fixes #235

### DIFF
--- a/fusesoc/capi2/core.py
+++ b/fusesoc/capi2/core.py
@@ -65,8 +65,10 @@ class String(str):
             "https://github.com/olofk/fusesoc/issues/new."
         )
 
+
 class Integer(int):
     pass
+
 
 class StringWithUseFlagsOrList:
     def __new__(cls, *args, **kwargs):
@@ -361,6 +363,10 @@ class Core:
             generators[k].root = self.files_root
             self._debug(" Found generator " + k)
         return generators
+
+    def get_virtuals(self):
+        """ Get a list of "virtual" VLNVs provided by this core. """
+        return self.virtual
 
     def get_parameters(self, flags={}, ext_parameters={}):
         def _parse_param_value(name, datatype, default):
@@ -710,6 +716,10 @@ Root:
     - name : vpi
       type : Vpi
       desc : Available VPI modules
+  lists:
+    - name : virtual
+      type : Vlnv
+      desc : VLNV of a virtual core provided by this core. Versions are currently not supported, only the VLN part is used.
 
 Fileset:
   description : A fileset represents a group of files with a common purpose. Each file in the fileset is required to have a file type and is allowed to have a logical_name which can be set for the whole fileset or individually for each file. A fileset can also have dependencies on other cores, specified in the depend section
--- a/fusesoc/coremanager.py
+++ b/fusesoc/coremanager.py
@@ -56,6 +56,13 @@ class CoreDB:
                 )
         return ", ".join(deps)
 
+    def _parse_virtual(self, virtuals):
+        package_names = []
+        for virtual in virtuals:
+            for simple in virtual.simpleVLNVs():
+                package_names.append("{}".format(self._package_name(simple)))
+        return ", ".join(package_names)
+
     def add(self, core, library):
         self._solver_cache_invalidate_all()
 
@@ -124,12 +131,22 @@ class CoreDB:
         cores = [x["core"] for x in self._cores.values()]
         for core in cores:
             if only_matching_vlnv:
-                if not eq_vln(core.name, top_core):
+                if not any(
+                    [eq_vln(core.name, top_core)]
+                    + [
+                        eq_vln(virtual_vlnv, top_core)
+                        for virtual_vlnv in core.get_virtuals()
+                    ]
+                ):
                     continue
 
             package_str = "{} {}-{}".format(
                 self._package_name(core.name), core.name.version, core.name.revision
             )
+            _virtuals = core.get_virtuals()
+            if _virtuals:
+                _s = "; provides ( {} )"
+                package_str += _s.format(self._parse_virtual(_virtuals))
             if not only_matching_vlnv:
                 _flags["is_toplevel"] = core.name == top_core
                 _depends = core.get_depends(_flags)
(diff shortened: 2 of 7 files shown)


imphil:
@martin-lueker This is the proposal for how to do "virtual cores" or "provides" or "interfaces" in FuseSoC. As suggested, I've moved away from the "provide" keyword and renamed it to "virtual". Do you think that addresses your concerns?

@olofk This is a concept we need in OpenTitan; I want to propose it here first to discuss about the syntax and the general concept before we introduce it in OT.

martin-lueker:
> @martin-lueker This is the proposal for how to do "virtual cores" or "provides" or "interfaces" in FuseSoC. As suggested, I've moved away from the "provide" keyword and renamed it to "virtual". Do you think that addresses your concerns?
> 
> @olofk This is a concept we need in OpenTitan; I want to propose it here first to discuss about the syntax and the general concept before we introduce it in OT.

That sounds reasonable to me.  Thanks.

olofk:
Let's go with this. Fire at will. Thanks for the work

olofk:
@imphil Should we just pull this one in? Want to have it before 2.0

PR #525, opened 2021-09-30
Small code cleanup in coremanager

imphil:


imphil:
That's only unused code and added comments that I stumbled across when doing other work.

imphil:
(And lint will fail until master is lint clean again.)

fusesoc/coremanager.py
```diff
@@ -144,8 +150,7 @@ def eq_vln(this, that):
             repo.add_package(package)
 
         request = Request()
-        simplevlnvs = top_core.simpleVLNVs()
-        for sv in simplevlnvs:
+        for _ in top_core.simpleVLNVs():
```
olofk, line 153: What on earth is this loop supposed to do? I have no memory of what my intentions were but it doesn't make sense now

olofk: @imphil Did we ever figure out what the hell this is?

imphil: Unclear indeed. Let me have a closer look in the coming days (forgot about it totally).

imphil: Looks like useless code indeed. I removed it.

imphil:
While at it I also added a second commit to fix a pytest deprecation.

olofk:
Looks perfectly fine. Thanks!

PR #532, opened 2021-12-10
Adding EDAM top_core key/value.

sjalloq:
This change is required to support the Edalize filelist backend.  I use it to get the path to the parent Core file so that I can write out a filelist using a relative path.

Discussion was ignored here: https://github.com/olofk/edalize/issues/25


olofk:
I don't think we can do that. EDAM is used by other tools that don't use core description files, so we can't break that abstraction.

sjalloq:
Not sure I understand, isn't the "toplevel" key using a core file?

```
        self.edalize = {
            "version": "0.2.1",
            "dependencies": {},
            "files": [],
            "hooks": {},
            "name": self.system_name or top_core.sanitized_name,
            "parameters": {},
            "tool_options": {},
            "toplevel": top_core.get_toplevel(self.flags),
            "top_core": top_core.core_file,
            "vpi": [],
        }
```

olofk:
The toplevel key is the toplevel module(s) of the design

sjalloq:
But in that snippet above, the toplevel module name is taken from the `top_core.get_toplevel()` method.  That was already in the codebase.  It's not my addition.  I just added the `top_core.core_file`.

I guess my point/question is, is `top_core` not a Core file?  It seems to come from a call to `self.core_manager.get_depends()`.  Or perhaps I'm just not understanding what you meant by, "we can't break that abstraction".

sjalloq:
Closing this PR and moving discussion back to https://github.com/olofk/edalize/issues/25

PR #533, opened 2021-12-10
feat_board_file_support: adding support for copying directories

sjalloq:
- copyto now supports copying directories to the workspace. This is useful
   for copying directories such as Xilinx board files for Vivado.

### DIFF
--- a/fusesoc/edalizer.py
+++ b/fusesoc/edalizer.py
@@ -192,7 +192,14 @@ class Edalizer:
                     _dstdir = os.path.dirname(dst)
                     if not os.path.exists(_dstdir):
                         os.makedirs(_dstdir)
-                    shutil.copy2(os.path.join(files_root, file["name"]), dst)
+                    try:
+                        shutil.copy2(os.path.join(files_root, file["name"]), dst)
+                    except IsADirectoryError:
+                        shutil.copytree(
+                            os.path.join(files_root, file["name"]),
+                            dst,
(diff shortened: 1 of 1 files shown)


sjalloq:
This PR is in support of https://github.com/olofk/edalize/pull/290

olofk:
How would you use this in a core description file? Do you specify the directory as `file_type : user`?

olofk:
Thanks. I see a drawback of this feature because it means that FuseSoC will not know _exactly_ which files are being copied to the build directory. OTOH, anything with file_type : user is ignored by the backends anyway and I can see that this is quite handy. Picked and pushed! Would be great with a test too if you feel for a follow-up patch

sjalloq:
Thanks, and yes, it was my plan to add a testcase.  However, I found a second point in the code where this needs to be merged in.  We've been running fusesoc with the `--no-export` switch.  However, if you run without it, the copy step happens in another section of code.

I'll push a fix for that now.

olofk:
Hmm... sounds like this could need some cleaning up in FuseSoC if there are two different code paths. Oh well. Looking forward to your patch so that I can see what's going on

sjalloq:
Sorry, thought this would auto-update but I guess it doesn't after it has been closed.  Here's the last fix:

https://github.com/sjalloq/fusesoc/commit/1a3ad9b751d0aaa4c5c21a7d31228dfc240e74f8

PR #536, opened 2021-12-23
Clarify need for fusesoc project directory

gojimmypi:
see also https://github.com/olofk/serv/pull/69

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,10 +38,12 @@ full list of system requirements and installation instructions in the
 To check if FuseSoC is working, and to get an initial feeling for how FuseSoC
 works, you can try to simulate a simple hardware design from our core libray.
 
-First, create and enter an empty workspace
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -38,10 +38,12 @@ full list of system requirements and installation instructions in the
 To check if FuseSoC is working, and to get an initial feeling for how FuseSoC
 works, you can try to simulate a simple hardware design from our core libray.
 
-First, create and enter an empty workspace
+The `fusesoc` install is specific to a project directory, and not a system-wide feature.
```
olofk, line 42: I think we should use some different wording here. FuseSoC itself is system-wide. The build-directory is workspace (project) specific and the libraries containing core files and fusesoc.conf can be both system-wide or workspace-specific depeending on whether it was added with the `--global` flag

olofk:

PR #545, opened 2022-02-10
Remove unused flags argument from Core.get_generators()

rswarbrick:
This argument doesn't have any effect, so there's no need to construct
it. This commit removes the argument from the function and all
call-sites.

It also removes debug messages from in `core.get_generators()`. The
first wasn't massively helpful (printing flags that had no effect),
and the interesting call site in `Edalizer.extract_generators()` already
has a log message which was more helpful because it could also print
the core name.

The commit moves the second debug message (per generator found) up to
`Edalizer.extract_generators`, where it was also duplicated.


rswarbrick:
I *think* that the CI failure is not caused by this change. Digging through the logs, I see this:
```
2022-02-10T13:54:35.7787514Z py3-ci run-test: commands[0] | /bin/rm -rf /home/runner/work/fusesoc/fusesoc/.tox/.tmp/homedir
2022-02-10T13:54:35.7951485Z /bin/rm: cannot remove '.tox/.tmp/homedir/.local/share/virtualenv/wheel': Directory not empty
2022-02-10T13:54:35.8090326Z ERROR: InvocationError for command /bin/rm -rf .tox/.tmp/homedir (exited with code 1)
```
but I don't think I have permissions to do anything cleverer. Any ideas?

imphil (approved):
LGTM. Yeah, the CI failure is very unlikely to be due to this change, let's ignore it for now :-)

imphil:
I haven't had a @rswarbrick PR in a while! :-)

PR #548, opened 2022-02-10
CI: Add Python 3.11 (alpha for now)

imphil:
Python 3.11 is in alpha now, but having it in CI helps us be prepared.
CI will always pick up the latest 3.11 version as it moves from alpha to
beta to release ultimately.


olofk:
Good idea. Let's test with Python for Workgroups

PR #549, opened 2022-02-10
Remove fusesoc pgm and fusesoc init

imphil:
These two commands have been deprecated for a while, remove them while
still showing a message to our users informing them what to do instead.


olofk:
Massive cleanup. My favorite! Picked and pushed. Thanks!

Issue #546, opened 2022-02-10
Spurious CI fails in the rm-rf line


imphil:
We get random CI fails when doing the "rm -r" part of the steps described in our tox config.

Here's one from macOS (https://github.com/olofk/fusesoc/runs/5144792316?check_suite_focus=true), but we've seen it on Linux as well:

```
   py3-ci installdeps: pytest, pytest-github-actions-annotate-failures
  py3-ci inst: /Users/runner/work/fusesoc/fusesoc/.tox/.tmp/package/1/fusesoc-0.1.dev1+g87ec9ae.zip
  py3-ci installed: attrs==21.4.0,edalize==0.3.3,fusesoc @ file:///Users/runner/work/fusesoc/fusesoc/.tox/.tmp/package/1/fusesoc-0.1.dev1%2Bg87ec9ae.zip,importlib-metadata==4.10.1,importlib-resources==5.4.0,iniconfig==1.1.1,Jinja2==3.0.3,jsonschema==4.4.0,MarkupSafe==2.0.1,okonomiyaki==1.3.2,packaging==21.3,pluggy==1.0.0,py==1.11.0,pyparsing==3.0.7,pyrsistent==0.18.1,pytest==7.0.0,pytest-github-actions-annotate-failures==0.1.6,PyYAML==6.0,simplesat==0.8.2,six==1.16.0,tomli==2.0.1,typing_extensions==4.0.1,zipfile2==0.0.12,zipp==3.7.0
  py3-ci run-test-pre: PYTHONHASHSEED='1614572130'
  py3-ci run-test: commands[0] | /bin/rm -rf /Users/runner/work/fusesoc/fusesoc/.tox/.tmp/homedir
  rm: .tox/.tmp/homedir/Library/Application Support/virtualenv/wheel/house: Directory not empty
  rm: .tox/.tmp/homedir/Library/Application Support/virtualenv/wheel: Directory not empty
  rm: .tox/.tmp/homedir/Library/Application Support/virtualenv: Directory not empty
  rm: .tox/.tmp/homedir/Library/Application Support: Directory not empty
  rm: .tox/.tmp/homedir/Library: Directory not empty
  rm: .tox/.tmp/homedir: Directory not empty
  ERROR: InvocationError for command /bin/rm -rf .tox/.tmp/homedir (exited with code 1)
```

This code was never anywhere close to beautiful, but now it also seems to be broken.

PR #550, opened 2022-02-10
Various cleanups

imphil:
Some things I stumbled across while debugging something else.


imphil:
The CI failure is #546

olofk:
Very nice. We will get CI to behave eventually. Thanks!

PR #553, opened 2022-02-17
feat_board_file_support: fix for missing copytree

sjalloq:
- there are two locations in the codebase where file copying takes place depending
   on whether you run fusesoc with the --no-export switch.


sjalloq:
@olofk This is the second half of #533.  I think you missed my last comment so I thought I'd just open a new PR.

olofk:
Yes. Totally missed that final comment. Thanks for the fix. Picked and pushed!

PR #555, opened 2022-03-10
Tweak test_edalizer not to depend on the order of edalizer.cores

rswarbrick:
This didn't matter until 0469da4 because all the 'core_root' paths
were the same. But generated cores now appear in different directories
from one another so we have to get it right.

Note that this isn't actually required for upstream at the moment: it's probably nice to be more robust, but things were working anyway. But it *is* required for some local changes we've got on our OpenTitan fork where @imphil's patch reorders generated cores a bit (and the "-3" and "-2" indices were then wrong).

### DIFF
--- a/tests/test_edalizer.py
+++ b/tests/test_edalizer.py
@@ -36,9 +36,12 @@ def test_generators():
     )
     edalizer.run()
 
-    gendir = edalizer.cores[-3].core_root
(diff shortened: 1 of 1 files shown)


olofk:
TBF I don't really understand the issue, but trust you on this one. Thanks!

PR #556, opened 2022-03-23
Enable Edalize flow API support

olofk:
This enables support for the new flow API in Edalize, which will eventually supersede the current Edalize API. For now these two APIs live side by side. FuseSoC users who wish to remain on the old Edalize API doesn't need to do anything. For users wanting to use the flow API, set the `flow` key in a target to one of the supported flows, and set the `flow_options` to control flow-specific options. These two keys have precedence over the `default_tool` and `tools` keys currently used in the target section.

This is also one of the two major changes I want to get in before the FuseSoC 2.0 release, so please take a look and comment accordingly

### DIFF
--- a/fusesoc/coremanager.py
+++ b/fusesoc/coremanager.py
@@ -254,6 +254,8 @@ class CoreManager:
                     except ImportError as e:
                         w = 'Failed to register "{}" due to unknown provider: {}'
                         logger.warning(w.format(core_file, str(e)))
+                    except ValueError as e:
+                        logger.warning(e)
         return found_cores
 
     def _detect_capi_version(self, core_file) -> int:
--- a/doc/source/user/build_system/flow_options.rst
+++ b/doc/source/user/build_system/flow_options.rst
@@ -1,0 +1,33 @@
+.. _ug_build_system_flow_options:
+
+Configuring a flow
+==================
+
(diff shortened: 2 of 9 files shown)


olofk:
Ping @imphil

fusesoc/coremanager.py
```diff
@@ -227,6 +227,8 @@ def find_cores(self, library):
                     except ImportError as e:
                         w = 'Failed to register "{}" due to unknown provider: {}'
                         logger.warning(w.format(core_file, str(e)))
+                    except ValueError as e:
+                        logger.warning(e)
```
imphil, line 231: When does that happen, and why is it harmless? A comment would be great here.

imphil (commented):
I have a hard time following how this API is intended to be used. Would you mind updating the documentation as well to explain how that feature is used?

* What is a flow? Which flows are available?
* What are flow options? Which options are available?

olofk:
You're right. Documentation is very much needed. Will start writing.

fusesoc/coremanager.py
olofk: Yeah.... about that... I haven't got a clue anymore what that was supposed to fix. Better to drop it until we figure it out

olofk:
Alright then. I have added some basic documentation. @imphil, could you take another look?

olofk:
I went ahead and merged this

PR #557, opened 2022-03-24
Fix interface incompatibility in test_virtual

imphil:
Commit 6b04028491c02308bec9423627c35b3456ea980a introduced new code
which broke a test. Fix that. (There was a mid-air collision of two
changes.)


olofk:
Cool. Thanks

PR #566, opened 2022-05-09
Add support for ignored directories to config files

rswarbrick:
The first 3 commits in this PR are minor cleanups. The most interesting (controversial?) bit is probably the first:

- **Bump to a working version of Black**: Although we've pinned Black to version 20.8b1, this doesn't pin the "click" library on which it depends. Bump Black to a version that works with the latest click.

After that, I made a couple of tidy-up commits in `config.py`. Then this commit might need a bit of thought:

- **Interpret relative paths in config file relative to it, not CWD**: This changes behaviour, but probably to something more useful than what we had before. I'd be very surprised if we had any users relying on the fact that relative paths in a config file were interpreted relative to CWD.

Finally, we can add the `ignored_dirs` key that motivated the whole thing.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,2 @@
+[flake8]
+max-line-length = 88
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -30,6 +30,6 @@ repos:
       - id: isort
         args: ["--profile", "black", "--filter-files"]
   - repo: https://github.com/psf/black
-    rev: 20.8b1
+    rev: 22.3.0
     hooks:
       - id: black
--- a/fusesoc/capi2/core.py
+++ b/fusesoc/capi2/core.py
@@ -49,7 +49,7 @@ class Any:
 
 
 class StringWithUseFlags(str):
-    """ A parsed string with support for use flags. """
+    """A parsed string with support for use flags."""
 
     def __init__(self, string):
         self.exprs = None
@@ -61,7 +61,7 @@ class StringWithUseFlags(str):
 
 
 class String(str):
-    """ A plain (unparsed) string. """
+    """A plain (unparsed) string."""
 
     def parse(self, flags):
         raise RuntimeError(
@@ -373,7 +373,7 @@ class Core:
         return generators
 
     def get_virtuals(self):
-        """ Get a list of "virtual" VLNVs provided by this core. """
+        """Get a list of "virtual" VLNVs provided by this core."""
         return self.virtual
 
     def get_parameters(self, flags={}, ext_parameters={}):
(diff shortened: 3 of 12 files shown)


rswarbrick:
Uh, that's not going to work: I think CI checks formatting for all files (not just the changed ones that the pre-commit hook checks). I'll spin out a separate PR that bumps Black to a version that works with the latest version of its dependencies and update all the code appropriately. I'll also drop the first commit from this PR and hope that everything still matches :-)

rswarbrick:
@olofk: Ping!

olofk (approved):
I have no idea what flake8 is but I trust this is a good thing

olofk (approved):
IIRC the file handle thing was added to support running unit tests, so if you found a better way to do that, I think it's all good

olofk (approved):

olofk (commented):
Not sure about this one tbh. I _think_ people put `.` as location in their libraries to always have cwd registered as a library. Or perhaps core libraries are not affected by this at all?

olofk (approved):
Perfect! Thanks

rswarbrick:
I'm not sure the GitHub review flow has quite worked here :-) I guess the "not sure about this one" comment is the change to use relative paths? If we don't want to break "." (I hadn't thought of that!), we could do something like add a syntax to mean "relative to this file". e.g. something like `<cfg_dir>/foo` or similar?

rswarbrick:
> I have no idea what flake8 is but I trust this is a good thing

Heh :-) It's a PEP8 checker which has reasonably good editor integration. (Including Emacs, which I use!) By default, it follows the original PEP8 guidelines that say to break at column 79. Apparently, Black's defaults are different (sigh), causing red squiggly lines everywhere!

rswarbrick:
> The changes are fine, but I just noticed that systems_root and cores_root have been marked as deprecated since February 2018 so I think we can safely drop these instead. We can either do that here and avoid creating the get_{systems,core}_root functions or as a follow-up patch later

Happy to do a follow-up to drop them (my inner lazy engineer reckons that this will be less work than ripping them out first and then rebasing this series :-p)

olofk:
*sigh* You're right that this review flow doesn't work all that well. Mostly because I take too long between reviews. But yes, I'm worried about a backwards-incompatible change when making paths relative fusesoc.conf rather than relative cwd. Having `.` in fusesoc.conf is the specific use-case I can think of. OTOH, it might very well be that the number of users who put` `.` in `fusesoc.conf`AND having `fusesoc.conf` in another dir than CWD are zero. It can also be argued that we haven't really made any promises. One compromise that I think could be very reasonable is to have paths in `fusesoc.conf`relative to that file, but when you specify `--cores-root=...` on the command-line, those are relative CWD. Perhaps that's already what's happening. If so, could you just confirm, and I think we can go with that

olofk:
Fuck it. Let's just pull this in and IF there are any issues (which there might very well not be) we deal with them. Thanks for your work. Picked and pushed!

Issue #530, opened 2021-11-18
Thread-safe generators


fantognazza:
Fusesoc is not thread-safe when processing generators: the same `.core` file is shared among different executions, and gets overwritten by independent fusesoc runs.

It is a real-case scenario with consistent reproducibility when spawning multiple fusesoc executions with GNU parallel command, even if independent `build-root` parameters are provided.

Here are reported two errors printed out by independent execution runs:

> INFO: Generating ${VENDOR}:${LIBRARY}:${NAME}-memory_generator:${VERSION}
> WARNING: The first line of the core file ${HOME}/.cache/fusesoc/generated/${VENDOR}_${LIBRARY}_${NAME}-memory_generator_${VERSION}/${NAME}-memory_generator.core must be "CAPI=1" or "CAPI=2:".  The first line of this core file is "".
> ERROR: Setup failed : Unknown file type

> INFO: Generating ${VENDOR}:${LIBRARY}:${NAME}-memory_generator:${VERSION}
> ERROR: Setup failed : Failed to parse generated core file ${NAME}-memory_generator.core: Missing 'name' parameter

Assuming that the called generate script itself is thread-safe, it would be desirable to achieve concurrent fusesoc runs containing generators.

olofk:
This is a problem indeed. I made a change now so that generators are run in unique temp directories instead. Please let me know if 0469da452a68d44689bcb347708c7f4a775d3e41 fixes the issue

fantognazza:
Looks good, the only minor issue I can see now is the generator's temporary directory persisting after the fusesoc command ends.

Calling the tool multiple times, like in the case of a design space exploration, leads to a `/tmp` folder full of orphan directories.
In fact, it can be considered somehow similar to a 'memory leak', being `/tmp` mapped to RAM memory.

olofk:
That's a fair point. We need to find some point where it's safe to remove them. Probably somewhere in the edalizer.run function

PR #568, opened 2022-05-25
Remove generators temporary directories

fantognazza:
This pull request completes the fix for #530.

Now temporary directories are removed at the end of their usage.

### DIFF
--- a/fusesoc/edalizer.py
+++ b/fusesoc/edalizer.py
@@ -88,6 +88,10 @@ class Edalizer:
 
         # Create EDA API file contents
         self.create_edam()
+
+        # Clean up ttptttg temporary directories
+        self.clean_temp_dirs()
+
         return self.edam
 
     def _core_flags(self, core):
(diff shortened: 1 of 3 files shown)


fusesoc/edalizer.py
```diff
@@ -240,6 +244,12 @@ def create_edam(self):
         for snippet in first_snippets + snippets + last_snippets:
             merge_dict(self.edam, snippet)
 
+    def clean_temp_dirs(self):
+        for core in self.cores:
+            if core.core_root.startswith(tempfile.gettempdir()):
+                logger.debug(f"Removing {core.core_root} ttptttg temporary directory")
```
olofk, line 250: Very clever!I'm just wondering if there could be any cases where the cores are stored in the temp dir and we don't want to remove them. Perhaps I'm being a bit paranoid here

fantognazza: We could keep track of the cores created by generators, and selectively remove only their `core_root`

olofk: That would be great if you think we can do it that way

PR #572, opened 2022-06-16
Adding support for no_export flag

shareefj:
This change adds support for setting a no_export flag from within a target which
prevents the source files from being exported to the build tree.  You can
currently get the same behaviour by passing the --no-export command line argument
but this relies on the user remembering to set it.

The use case is for flows that would break if the files were copied to the build
tree, such as linting where the waivers reference absolute paths.


olofk:

Issue #580, opened 2022-08-01
Bug: Fusesoc gives error when unrelated core file from llvm is in the sub directories


hossein1387:
I am trying to add Fusesoc to [Ara](https://github.com/pulp-platform/ara). One of the dependency of this project is risc-v LLVM which is populated when the project is cloned. However, calling any Fusesoc command that reads a core file will fail if issued in the top ara directory. 

#### How to reproduce the bug (warning: it takes a long time to clone risc-v gcc and llvm repositories):

      git clone https://github.com/pulp-platform/ara.git
      git submodule update --init --recursive
      git submodule sync --recursive
      touch ara.core
      fusesoc library add ara .
      fusesoc core list

The last command produces this error:

```
Traceback (most recent call last):
  File "/export/tmp/hemmat/anaconda3/bin/fusesoc", line 8, in <module>
    sys.exit(main())
  File "/export/tmp/hemmat/anaconda3/lib/python3.9/site-packages/fusesoc/main.py", line 774, in main
    cm = init_coremanager(config, args.cores_root)
  File "/export/tmp/hemmat/anaconda3/lib/python3.9/site-packages/fusesoc/main.py", line 521, in init_coremanager
    cm.add_library(library)
  File "/export/tmp/hemmat/anaconda3/lib/python3.9/site-packages/fusesoc/coremanager.py", line 233, in add_library
    self._load_cores(library)
  File "/export/tmp/hemmat/anaconda3/lib/python3.9/site-packages/fusesoc/coremanager.py", line 220, in _load_cores
    found_cores = self.find_cores(library)
  File "/export/tmp/hemmat/anaconda3/lib/python3.9/site-packages/fusesoc/coremanager.py", line 209, in find_cores
    core = Core(core_file, self.config.cache_root)
  File "/export/tmp/hemmat/anaconda3/lib/python3.9/site-packages/fusesoc/core.py", line 17, in __new__
    l = f.readline().split()
  File "/export/tmp/hemmat/anaconda3/lib/python3.9/codecs.py", line 322, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xa8 in position 72: invalid start byte
```     

[...]

PR #581, opened 2022-08-02
Catch exception when a core file is not in text format

hossein1387:
This is a potential fix for issue https://github.com/olofk/fusesoc/issues/580. When a core file not in text format is passed to be checked for CAPI version, there is no code to handle exceptions that are raised after opening the file. As explained in issue https://github.com/olofk/fusesoc/issues/580, an instance of such scenario is when `riscv-llvm` repository exist in the current directory, FuseSoc tries to read `riscv-llvm/lldb/test/Shell/Register/Core/Inputs/x86-64-netbsd.core` which is a binary file. This PR wraps the `open(core_file)` function and ignores the file and provides a warning to the user that which files were ignored. The user, as @olofk suggested, should consider ignoring the parent directory as described [here](https://github.com/olofk/fusesoc/pull/566).

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,2 @@
+[flake8]
+max-line-length = 88
--- a/fusesoc/capi2/core.py
+++ b/fusesoc/capi2/core.py
@@ -53,7 +53,7 @@ class Any:
 
 
 class StringWithUseFlags(str):
-    """ A parsed string with support for use flags. """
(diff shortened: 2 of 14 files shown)


olofk:
This looks perfectly fine to me. Picked and pushed. I had to do some manual conflict resolving though, so please check that I got it right. Thanks for your contribution

PR #582, opened 2022-08-03
Integrate openTitan `virtual` keyword (proposal)

joennlae:
@imphil I tried to rebase this such that this addition is integrated into the current version of `fusesoc`. I got stuck during rebasing as I do not have enough experience with all the code. Is it possible to rebase that into the master branch here 👍 ? If that could be rebased, one does not have to use this (`lowRISC/ot`) branch to integrate `lowRISC` IP.


imphil:
The ot branch contains various things.
* Support for the virtual keyword. That's already merged in upstream FuseSoC at https://github.com/olofk/fusesoc/pull/524.
* Support for dependency resolution in generators. That's https://github.com/olofk/fusesoc/pull/391, and has significant design problems which require this feature to be redesigned. It works for OKish OpenTitan, but it's not ready for general use.

joennlae:
Thank you for clarifying 👍

PR #583, opened 2022-08-05
Add CLI for listing tool backends

olofk:
This adds `fusesoc tool list` to the CLI. A FuseSoC `tool show <tool>` should be added as well but might benefit from some refactoring first to get nice option printing like when running `fusesoc run <core> --help`.

Also, once we pull in the flow API, corresponding `fusesoc flow show/list` will be added as well

### DIFF
--- a/fusesoc/main.py
+++ b/fusesoc/main.py
@@ -198,6 +198,22 @@ def list_cores(cm, args):
         )
 
 
+def list_tools(cm, args):
+    from edalize.edatool import walk_tool_packages
+
+    _tp = list(walk_tool_packages())
+    maxlen = max(map(len, _tp))
+
+    for tool_name in _tp:
+        try:
+            tool_class = get_edatool(tool_name)
+            desc = tool_class.get_doc(0)["description"]
+            print(f"{tool_name:{maxlen}} : {desc}")
+        # Ignore any misbehaving backends
+        except Exception:
+            pass
+
+
 def gen_list(cm, args):
     cores = cm.get_generators()
     if not cores:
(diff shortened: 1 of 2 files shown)


fusesoc/main.py
```diff
@@ -198,6 +198,18 @@ def list_cores(cm, args):
...
+def list_tools(cm, args):
+    from edalize.edatool import walk_tool_packages
+
+    _tp = list(walk_tool_packages())
+    maxlen = max(map(len, _tp))
+
+    for tool_name in _tp:
+        tool_class = get_edatool(tool_name)
+        desc = tool_class.get_doc(0)["description"]
+        print(f"{tool_name:{maxlen}} : {desc}")
```
imphil, line 210:
```suggestion
        print(f"{tool_name:{maxlen}}: {desc}")
```

fusesoc/main.py
```diff
@@ -528,6 +540,17 @@ def get_parser():
...
+    tool_subparsers = parser_tool.add_subparsers()
+    parser_tool.set_defaults(subparser=parser_tool)
+
+    # tool list subparser
+    parser_tool_list = tool_subparsers.add_parser("list", help="List available tools")
+    parser_tool_list.set_defaults(func=list_tools)
+
     # list-cores subparser
     parser_list_cores = subparsers.add_parser("list-cores", help="List available cores")
     parser_list_cores.set_defaults(func=list_cores)
```
imphil, line 556: Looks like this command should then be "core list" as well in a follow-up?

imphil (approved):
LGTM. How about adding a test as well?

fusesoc/main.py
olofk: I wanted to have it like `fusesoc core list` which has a space between the longest name and the `:`

fusesoc/main.py
olofk: `list-cores` and `core-info`have been superseded by `core list` and `core show` since 8284202ed54e0ca0b6e57b6cddffa078b1aa4c9b Planning to remove the old commands at some points. Doing it at 2.0 would had been nice but we will need to do it later since we haven't put in a deprecation notice yet

olofk:
A test would be nice bu I'm not sure how to deal with that. We both require (a mocked version of?) Edalize and calling the FuseSoC command-line, so I think I'll skip that for now and focus on other things.

I did however realize I need to encapsulate the backend introspection in a try/catch because with upstream Edalize it will fail otherwise. Amending and force-pushing

Issue #576, opened 2022-07-07
Broken "version" attribute in git provider


JLeemaster:
### Summary
Specifying a `version`  attribute when using the `git` provider is currently broken as implemented. the root cause was introduced by commit 8911f7c.

The current implementation will perform `git clone` with a `--depth 1`  argument.  A side effect of cloning with the `--depth 1` flag is that all tags & branch references that are not the current `HEAD`  will be omitted from the cloned repo.  Therefore, if the git provider's `version` attribute is set to any tag or branch other than the current `HEAD` of the repository, it will fail when trying to checkout the branch or tag specified in the `version` attribute.

### How to reproduce

1. Clear the fusesoc cache
2. Take any known-good `.core` file that uses the `github` provider and specifies a version that is not the most recent `master` or `main` commit for that repository
3. Convert the `.core` file to use the `git` provider instead of the `github` provider
4. Attempt to use it with fusesoc to download the core

### Suggested solution

_After_ the git provider executes the `git clone --depth 1` command:

1. Retrieve the list of available references from the remote repo
   - e.g. `git ls-remote origin`
2.  Scan the returned list for a branch or tag that matches `version`
3. If a match is found, or if `version` is a fully-specified SHA1 commit ID, fetch it accordingly
      - if it's a tag:
         - fetch it and make it a local tag
            - `git fetch --depth 1 origin refs/tags/${version}:refs/tags/${version}`
      - if it's a branch _or_ a fully-specified SHA1 commit ID:
         - fetch it normally
            - `git fetch --depth 1 origin ${version}`
5. checkout the fetched branch, tag, or commit ID
   - If it is a branch or a tag:
      - `git checkout ${version}`
   - if it is a full SHA1 commit ID:
      - `git checkout FETCH_HEAD // detached-head checkout of whatever was fetched`

This will keep the benefits of doing a shallow `git clone` while still being able to specify branches, tags, & commit IDs that are not the current `HEAD` of the remote repo as the core's `version`.

mithro:
Maybe the stuff at https://github.com/YosysHQ/yosys/pull/3138 could help?

GCHQDeveloper211:
Hi @olofk,

We have just found this issue too, and came to a solution before seeing this ticket.

[...]

PR #585, opened 2022-08-23
Make sure shallow clone pulls all tags

GCHQDeveloper147:
No-single-branch flag ensures shallow-clone depth constraint does not make older tags invisible

(Issue #576)


imphil (approved):
LGTM, thanks!

PR #587, opened 2022-09-21
Docs: Installing fusesoc under Windows

orkaboy:
Installation instructions for python/fusesoc on Windows added to user guide.

### DIFF
--- a/doc/source/user/installation.rst
+++ b/doc/source/user/installation.rst
@@ -97,7 +97,7 @@ If an older version of FuseSoC is found, this version is upgraded to the latest 
 
 .. code-block:: shell-session
 
-   $ pip3 install --upgrade --user fusesoc
+   $ pip3 install --upgrade fusesoc
 
 Check that the installation worked by running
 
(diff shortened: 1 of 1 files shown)


orkaboy:
Verified on Windows 10 using python 3.10.7. **fusesoc** is correctly installed by **pip3**, and running `fusesoc --version` returns `1.12.0`.

imphil (commented):
Nice, thank you! Some small nits, but then we're ready to go.

To make the installation steps self-contained, how about adding the "pip install" steps to the instructions as well, instead of relying on the Linux instructions?

orkaboy:
I applied your fixes and copy-pasted the text from Linux, removing any Linux specific stuff such as sudo.

The one thing I'm not sure of is if the installation is actually "system-wide". There is a difference when using the `--user` flag; the script is installed to `Users/youruser/AppData/Roaming` as opposed to `Users/youruser/AppData/Local/Programs`. In both cases it will be installed under the current user though.

doc/source/user/installation.rst
```diff
@@ -76,9 +76,56 @@ Use ``pip`` to remove FuseSoC from your system.
...
+
+Check that the installation worked by running
+
+.. command-output:: fusesoc --version
 
-    Add Windows installation instructions.
+
+If this command works FuseSoC is installed properly and ready to be used.
+
+If the shell reports an error about the command not being found verify that the ``Scripts`` directory (example: ``C:/Users/youruser/AppData/Roaming/Python/Python310/Scripts``) folder is added to the search path (``PATH``).
```
imphil, line 109: Is "fusesoc" installed into the Scripts folder? Or do you need to add another path to PATH?

doc/source/user/installation.rst
```diff
@@ -76,9 +76,56 @@ Use ``pip`` to remove FuseSoC from your system.
...
+
+
+System-wide installation
+------------------------
+
+FuseSoC can be installed for all users on a system.
+
+.. code-block:: shell-session
+
+   $ pip3 install --upgrade fusesoc
```
imphil, line 119: As you mentioned, both installations are "per user". So simply remove the user installation instructions above, keeping only the "system ones" (and make it the one and only instructions).

doc/source/user/installation.rst
orkaboy: Yes, it's installed in Scripts (the `AppData/Local/Programs/Python/Python310/Scripts` folder is automatically added to PATH when installing Python):
image

doc/source/user/installation.rst
orkaboy: Will fix.

imphil (approved):
Looks good, thanks for your contribution!

PR #593, opened 2022-11-21
Vunit dependencies

nopeslide:
added fields to return value of get_files
  - "fileset": fileset which included the src file
  - "core": name of the core which included the src file
  - "depend": name list of core dependencies

### DIFF
--- a/fusesoc/capi2/core.py
+++ b/fusesoc/capi2/core.py
@@ -261,6 +261,7 @@ class Core:
         for f in src_files:
             for filename, attributes in f.items():
                 attributes["name"] = filename
+                attributes["core"] = self.name
 
                 # Remove all key-value-pairs with values that are either bool with
                 # value False or str of length 0


nopeslide:
WIP because I do not know if this is the intended way to add such a feature

olofk:
Ok, this is an interesting feature and I'm definitely open to adding more metadata to the EDAM file. I just want to be careful to not expose too much internal FuseSoC details since Edalize is intended to work without FuseSoC (and it is already used by a bunch of other projects).

Adding core is the least controversial in my opinion.

I'm less sure about fileset. To me that feels like an internal FuseSoC detail. Do you need that? I don't see it being used in https://github.com/olofk/edalize/pull/358

I see two issues with depend. The first is that we probably need to parse the list to evaluate the flags before sending it to the EDAM. The other thing is that the EDAM file already contains the dependency tree, so if you know which core a file belongs to, you can figure out its dependencies. So with that, I don't think we need to specify dependencies on a file-level as well.

So, I'm happy to accept a modified patch that only adds the "core" field. If you still want to include the other stuff, then we need to discuss that.

nopeslide:
@olofk 
I didn't know the edam already contained the dependency tree

nopeslide:
Anything else I missed?
I did no thorough testing, is there any test setup I can use?

olofk:
I tested this just now and there is actually a test that breaks with this change. test_capi2_get_files in `https://github.com/olofk/fusesoc/blob/main/tests/test_capi2.py` breaks. Also, I think you should convert the Vlnv object to a string first, instead of adding it directly

olofk:
@nopeslide I'm planning to release FuseSoC 2.0 very soon. If we want this patch to go in before that, then we need to fix the failing test and do the VLNV to string conversion.

nopeslide:
@olofk rebased on latest main
see olofk/edalize#384

olofk:
This completely slipped my mind, and I also realized that this is actually implemented since some time ago, so I'll thank you for the contributions and closing this now.

PR #594, opened 2022-12-01
Adding support for no_export flag

shareefj:
This change adds support for setting a no_export flag from within a target which prevents the source files from being exported to the build tree.  You can currently get the same behaviour by passing the --no-export command line argument but this relies on the user remembering to set it.

The use case is for flows that would break if the files were copied to the build tree, such as linting where the waivers reference absolute paths.

@olofk This was a PR you closed by mistake and I've rebased it on latest and re-opened.

### DIFF
--- a/fusesoc/capi2/core.py
+++ b/fusesoc/capi2/core.py
@@ -322,6 +322,8 @@ class Core:
             # Special case for tool as we get it from default_tool
             if target.default_tool:
                 flags["tool"] = str(target.default_tool)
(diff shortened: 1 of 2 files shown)


shareefj:
@olofk Actually, this seems to be broken following the type checking changes.  How are we now supposed to use boolean values in core files?

shareefj:
Meh, looks like this is well out of date now. I can just use `flags` directly in a target.

PR #599, opened 2022-12-22
Enable use flags in tool options

Jbalkind:
At present, tool options aren't being parsed for use flags. As far as I can tell from my testing, this one line change should enable use flags within tool options. I realise this may not be the current behaviour so it's a breaking change, but I think it's necessary for some of our use cases and it seems unlikely that there are a lot of tool options currently in use that look like use flags.

Happy to discuss alternatives to this if it's not suitable for merging.


olofk:
This looks perfectly fine to me. Many thanks!

PR #600, opened 2022-12-23
[RFC] Add early environment variable resolution to fusesoc

Jbalkind:
Adds a --resolve-env-vars-early flag to FuseSoC which allows for environment variables in core files. These are resolved in FuseSoC rather than edalize. I imagine we could add a --resolve-env-vars-late (or make this into a flag with an argument) to enable resolution in edalize for use cases that need that.

### DIFF
--- a/fusesoc/capi2/core.py
+++ b/fusesoc/capi2/core.py
@@ -179,18 +179,22 @@ class Provider:
 class Core:
     capi_version = 2
 
-    def __init__(self, core_file, cache_root="", generated=False):
+    def __init__(
+        self, core_file, cache_root="", generated=False, resolve_env_vars=False
+    ):
         self.core_file = core_file
 
         basename = os.path.basename(self.core_file)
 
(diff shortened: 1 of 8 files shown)


olofk:
This is fine with me. It looks like macosx tests are broken though. Could it be that mkdir is not in /usr/bin on macosx? If so, do we really need to specify an absolute path?

olofk:
Ah! I understand now that all tests were broken prior to this. Looking at the tox docs, it seems perfectly fine to just allowlist the command without an absolute path, so could you please amend your patch to remove `/usr/bin/' to make tests work again on MacOSX as well?

Jbalkind:
Ok between that and pulling in your opencores patch, the CI passes. I can juggle those commits around in some other way or you can handle it since I know you don't like merges.

olofk:
Many thanks! All good from my side. I can do the cherry-picking.

olofk:
Picked and pushed! All back to green. Thanks

PR #607, opened 2023-02-16
Fall back if shallow clones fail

### LINKED ISSUE #606: Some remotes do not support shallow clones
I've recently upgraded to v2.0, and most things are working a charm. One thing has stopped me in my tracks, though, and that's that one of our remote git repos does not support shallow clones. Fusesoc 2 seems to attempt a shallow clone regardless for git, which means any cores that point to remote repositories with the git provider that do not support shallow cloning fail.

The idea of the shallow clone is definitely a good one, particularly for larger repos. I'll submit a simple patch that is working for me that tries to do a shallow clone, but if that fails falls back to trying without.

GCHQDeveloper137:
If an attempt to shallow clone a git repo fails, try a non-shallow clone, as this feature is not supported by all remotes.

fixes #606

### DIFF
--- a/fusesoc/provider/git.py
+++ b/fusesoc/provider/git.py
@@ -37,8 +37,12 @@ class Git(Provider):
         # TODO : Sanitize URL
         repo = self.config.get("repo")
         logger.info("Checking out " + repo + " to " + local_dir)
-        args = ["clone", "-q", "--depth", "1", "--no-single-branch", repo, local_dir]
-        Launcher("git", args).run()
+        try:
+            args = ["clone", "-q", "--depth", "1", "--no-single-branch", repo, local_dir]
+            Launcher("git", args).run()
+        except:
+            args = ["clone", "-q", "--no-single-branch", repo, local_dir]
+            Launcher("git", args).run()
         if version:
             args = ["-C", local_dir, "checkout", "-q", version]
             Launcher("git", args).run()
--- a/fusesoc/provider/git.py
+++ b/fusesoc/provider/git.py
@@ -40,9 +40,12 @@ class Git(Provider):
         try:
             args = ["clone", "-q", "--depth", "1", "--no-single-branch", repo, local_dir]
             Launcher("git", args).run()
-        except:
-            args = ["clone", "-q", "--no-single-branch", repo, local_dir]
-            Launcher("git", args).run()
+        except RuntimeError as e:
+            try:
+                args = ["clone", "-q", "--no-single-branch", repo, local_dir]
+                Launcher("git", args).run()
+            except:
+                raise e
         if version:
             args = ["-C", local_dir, "checkout", "-q", version]
             Launcher("git", args).run()


olofk:
Oh. I had no idea that shallow clones was an optional feature. git never ceases to surprise me. Yes, this feature makes sense. I also guess this solves your problem, but before merging I would like to address two things.
1. Can we test for shallow support instead of trying and failing? Looking around a bit I see indications that modern git servers might expose their capabilities but I didn't get a good idea of how reliable this is. I also saw someone claiming that trying to do a shallow clone just hangs, and in that case it doesn't help to retry if the first command never completes. Not sure if that was a very old git though
2. Is there some implication we are missing by catching every exception in the try/except clause. Will this cause us to report the wrong error somewhere?

olofk:
> > 1. Can we test for shallow support instead of trying and failing? Looking around a bit I see indications that modern git servers might expose their capabilities but I didn't get a good idea of how reliable this is. I also saw someone claiming that trying to do a shallow clone just hangs, and in that case it doesn't help to retry if the first command never completes. Not sure if that was a very old git though
> 
> There's a very good question indeed. I'm not an expert at all, but have done some digging around and this doesn't seem to be something that's readily exposed by git. I've seen various examples of using `get fetch -v -v` and looking for "Server supports" or similar, but people have reported both false positives and false negatives using this method. I've done some tests, and have certainly found remotes that don't advertise shallow capabilities, but seem to allow them. If anyone has any better/more reliable ideas, I'd love to hear them and would be very willing to test them out and update the PR accordingly.
> 
> The hanging is more of an issue for sure - a second try certainly isn't going to help there, of course. That's not something I've experienced myself at all. I wonder if that might be a symptom of a different problem?
> 

That pretty much mirrors my results as well. Let's go with your solution and we'll just be aware that we might need to refine it if someone gets into the hanging situation

> > 2. Is there some implication we are missing by catching every exception in the try/except clause. Will this cause us to report the wrong error somewhere?
> 
> That's a possibility I'd not considered at all! My python skills are really quite basic, most of my career has been spent on VHDL and kernel-level C. I suppose we can look at raising the original exception if the retry fails, to preserve the existing behaviour? I'll try to get an update in to fix this.

Please take another look and see if you can improve it somehow. I'm thinking for example of a situation where the `git` command doesn't exist at all. Perhaps you could single out and just catch whatever error that git throws when a shallow clone doesn't work.

GCHQDeveloper137:
> Please take another look and see if you can improve it somehow. I'm thinking for example of a situation where the `git` command doesn't exist at all. Perhaps you could single out and just catch whatever error that git throws when a shallow clone doesn't work.

It seems we were both thinking along the same lines there. I've done exactly that, and it seems to work with some local testing. If you think it needs any further adjusting, feel free to give me a nudge and I'll get on to it.

olofk:
I think this is fine now. Thank you for the contribution. Picked and pushed!

PR #608, opened 2023-02-28
Adding support for no_export flag.

shareefj:
Replicate --no-export command line argument using flags. This is useful for tool flows such as filelist generation where you don't want to copy all source files to the Fusesoc build directory.


shareefj:
@olofk This was equivalent to an old PR I created, you closed during a clean up, I re-opened and subsequently closed after convincing myself that there was already support for this when there wasn't.  So here it is again.

The use case is really just to enable specific tool flows to prevent source files being exported without having to rely on the user to remember to supply the command line argument.  I currently only use this in filelist generation.

PR #610, opened 2023-03-03
Allow double quotes in string expressions

shareefj:
Passing tool arguments such as -CFLAGS "blah blah" requires '"'.


olofk:
Great. Thanks. I found a similar issue with backticks just last week. We will probably hit a few more of these :/

PR #611, opened 2023-03-07
Modified yaml_fwrite to preserve file insertion order

leoechevarria:
Added `sort_keys=False` to `yaml.dump` call in `yaml_fwrite` to preserve the order of insertion of the files to be written in the .core when writing a generator (for instance for packages to be included and compiled before other sources). Supported from PyYAML >= 5.1.


imphil:
Thanks for your contribution! Can you add a test to show the problem you're fixing, and to ensure that it actually stays fixed?

And please add a version constraint on pyyaml if you depend on a feature that's only available from a certain version on (https://github.com/olofk/fusesoc/blob/main/setup.py#L56).

PR #636, opened 2023-07-31
Prepending generated core files

garankonic:
This merge request allows to prepend generated core files right before the core that instantiates a generator. 

Use-case: ModelSim requires vhdl files to be built in the right order. Imagine you generate a vhdl file (e.g. package with settings) and you want to use it already in the parent core. Using "first" position does not work if the package itself depends on some other package (core). 
Here is the example dependency tree, to make things clear:
common_core <- generated_core (depends on  common_core) <- current_core (has common_core in dependencies, instantiates generator to produce generated_core). 
If using "first" keyword, it would appear in the following order "generated_core <- common_core <- current_core" - which will not work.

### DIFF
--- a/fusesoc/capi2/json_schema.py
+++ b/fusesoc/capi2/json_schema.py
@@ -289,9 +289,9 @@ capi2_schema = """
               "type": "string"
             },
             "position": {
-              "description": "Where to insert the generated core. Legal values are *first*, *append* or *last*. *append* will insert core after the core that called the generator",
+              "description": "Where to insert the generated core. Legal values are *first*, *prepend*, *append* or *last*. *prepend* (*append*) will insert core before (after) the core that called the generator",
               "type": "string",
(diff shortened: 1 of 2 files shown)


olofk:
Excellent! Thanks for fixing this. It doesn't pass the syntax check however. [Here](https://fusesoc.readthedocs.io/en/stable/dev/devsetup.html#setup-development-environment) are some quick notes on how to enable the automatic syntax checks when you commit code.

Bonus points if you can come up with a testcase for this as well so that we don't accidentally break it in the future

garankonic:
Hey. Sorry for reacting so late, I thought adding tests would take more time:) Now should work fine (apart from the readthedocs - seems that your latest commit to the main affected this job somehow)
Cheers

olofk:
Excellent. Looks perfectly fine now. Thank you for your contribution. Picked and pushed!

PR #643, opened 2023-08-31
Add core filename to output of `fusesoc core show`.

shareefj:
This commit updates the output of `fusesoc core show` to add the filename of the core file itself. This change is required for tools such as the filelist Edalize backend as discussed in
https://github.com/olofk/edalize/issues/25.

The output of `fusesoc core show` now looks like:

CORE INFO
Name:        vee:el:en:vee
Description: Top bombing
Core root:   /root/path/to/core/file
Core file:   basename.core

### DIFF
--- a/fusesoc/capi2/core.py
+++ b/fusesoc/capi2/core.py
@@ -33,8 +33,7 @@ class Core:
 
         self.cache_root = cache_root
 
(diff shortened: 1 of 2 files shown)


olofk:
Thanks. This is very useful. Picked and pushed!

PR #644, opened 2023-09-03
Fixed setup dependency versions

sifferman:
I tested FuseSoC with all versions of pyparsing, pyyaml, fastjsonschema, jsonschema2md, and myst_parser. I found and added the minimum required versions. (fastjsonschema and jsonschema2md can be any version)

This fixes a setup issue I was having

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -52,12 +52,12 @@ setup(
     ],
     install_requires=[
         "edalize>=0.4.1",
-        "pyparsing",
-        "pyyaml",
+        "pyparsing>=2.3.1",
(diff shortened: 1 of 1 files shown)


olofk:
Many thanks. Picked and pushed!

olofk:
Ok, so I found one issue. myst_parser>=0.18.0 doesn't exist for Python 3.6. Now, the question is if FuseSoC with Python 3.6 has been broken for some time, or if we can safely lower the minimum version requirement for myst_parser

PR #645, opened 2023-09-03
Improved inheritance

### LINKED ISSUE #624: parameters_append overwrite inherited parameters_append
Hi! First of all, thanks for this awesome tool! It's reducing a lot of friction in my day-to day development, and being able to finally have a sane way of setting up continuous integration is priceless :D

I'm having an issue that I'm not sure if it's the expected behavior. When I use ``parameters_append`` more than once, the previous inherited ``parameters_append`` properties get overwritten by the last one.

In the MWE below, target ``default`` defines ``PARAM1`` (with ``parameters``), target ``child`` (which inherits from ``default``) defines ``PARAM2`` (with ``parameters_append``)  and target ``grandchild`` (which inherits from ``child``) defines ``PARAM3`` (with ``parameters_append``). When setting up target ``grandchild``, ``PARAM2`` doesn't appear in the generated tcl scripts.

Since using ``parameters`` in a target would already override all inherited parameters, I would expect ``parameters_append`` to allow appending a number of parameters among different inheritance steps.

I suppose the overwriting is happening at the YAML level, nevertheless it would be nice to have a way of having multiple ``parameters_append``, to avoid having duplicated constants.

## How to reproduce

1. Copy the following YAML code into a file named ``test_parameters_append.core``

```YAML
CAPI=2:                                                                         
name: ::test_parameters_append:0.1                                              
description: Testing whether a second parameters_append overwrites the first one
                                                                                
targets:                                                                        
  default: &default                                                             
    description: default target, includes "parameters" property                 
    default_tool: ise                                                           
    toplevel: this_entity_doesnt_exist                                          
    parameters: [PARAM1=1]                                                      
                                                                                
  child: &child                                                                 
    <<: *default                                                                
    description: inherits from default, has "parameters_append" property        
    tools:                                                                      
      ise:                                                                      
       family: Spartan6                                                         
       device: xc6slx9                                                          
       package: csg324                                                          
       speed: -2                                                                
    parameters_append: [PARAM2=2]                                               
                                                                                
  grandchild:                                                                   
    <<: *child                                                                  
    description: inherits from child, has "parameters_append" property          
    parameters_append: [PARAM3=3]                                               
                                                                                
parameters:                                                                     
  PARAM1:                                                                       
    datatype    : int                                                           
    description : First parameter                                               
    paramtype   : generic                                                       
  PARAM2:                                                                       
    datatype    : int                                                           
    description : Second parameter                                              
    paramtype   : generic                                                       
  PARAM3:                                                                       
    datatype    : int                                                           
    description : Third parameter                                               
    paramtype   : generic              
```

2. Have Xilinx ISE in your ``$PATH`` (the issue also happens with Vivado, so Vivado could also be used)

3. Run the command ``fusesoc --verbose --cores-root . run --setup --target=grandchild ::test_parameters_append:0.1``

## What I expected to happen

I expected the generated  tcl file at ``build/test_parameters_append_0.1/grandchild-ise/test_parameters_append_0.1.tcl`` to contain the following line:

```tcl
project set "Generics, Parameters" "PARAM1=1|PARAM2=2|PARAM3=3" -process "Synthesize - XST"
```
## What actually happened

Actually, ``PARAM2`` is missing from the generated tcl file:

```tcl
project set "Generics, Parameters" "PARAM1=1|PARAM3=3" -process "Synthesize - XST"
```

### LINKED ISSUE #639: Issues with Merge Keys
Hello! I have been using FuseSoC and Edalize for a few years now, and I've run into a lot of issues with the merge key feature from YAML 1.1. I wanted to open a discussion about examples, workarounds, and possible permanent fixes.

## `_append` strange behavior

If a child and grandchild both have a field with `_append`, then the grandchild overwrites the child's field. I would expect that both `_append`s are added to the parent's field.

This is already described in #624

### Example

```yaml
targets:
  default: &default
    filesets: [rtl1]

  child: &child
    <<: *default
    description: inherits from default
    filesets_append: [rtl2]

  grandchild:
    <<: *child
    description: inherits from child
    parameters_append: [rtl3] # overwrites rtl2
```

### Workaround

```yaml
targets:
  default: &default
    filesets: [rtl1]

  child: &child
    <<: *default
    description: inherits from default
    filesets_append: [rtl2]

  grandchild:
    <<: *child
    description: inherits from child
    parameters_append: [rtl2, rtl3] # must insert both
```

## Subfields are not copied

If a key exists in a child that matches a key from the parent, none of the subfields from the parent are given to the child. I would expect that the parent fields would be treated as defaults, then the child overwrites what it wants.

### Example

```yaml
targets:
  default: &default
    tools:
      verilator:
        mode: cc
        verilator_options:
          - --timing
  child:
    <<: *default
    tools:
      verilator:
        mode: lint-only
        # does not inherit verilator_options
```

### Workaround

```yaml
targets:
  default: &default
    tools:
      verilator:  &verilator
        mode: cc
        verilator_options:
          - --timing
  child:
    <<: *default
    tools:
      verilator:
        <<: *verilator # grab the subfield directly
        mode: lint-only
```

## Possible Solution

From these examples, it becomes clear that the merge key operation has some flaws.

One solution may be to create a new FuseSoC-only operator, (maybe `$<`), that fixes the above issues. I would be happy to add this functionality if people feel it would be valuable. I imagine that the best place to handle this is in CoreParser.read():

https://github.com/olofk/fusesoc/blob/40ca3f4765897c969c727dac1adba865d2c7148d/fusesoc/parser/coreparser.py#L49-L55

Thoughts?

sifferman:
Fixes #639 and fixes #624

In summary, inheritance is no longer implemented by PyYAML's merge key implementation. Instead, it is overridden by a custom FuseSoC implementation

## Specifics

* Added [fusesoc/capi2/inheritance.py](https://github.com/sifferman/fusesoc/blob/9af5a0ef42d96a8761090cc0b5abbe0c379340da/fusesoc/capi2/inheritance.py) to centralize logic related to inheritance

* `utils.yaml_fread()` now calls `utils.yaml_read()` to improve code reusability

* <ins>**Possibly**</ins> fixed small bug in `utils.yaml_read()` by changing `data.read()` to `data`. (There is no test in the CI that checks for this case, so I am uncertain whether this change is good)

* `utils.yaml_read()` now replaces all merge key operators (`<<`) with an intermediate value of `Inheritance.MERGE_OPERATOR`, then turns the YAML into a dictionary as usual, then runs the new `Inheritance.elaborate_inheritance()` function that matches the specifications in #639 

* Gave `utils.merge_dict()` the flag `concat_list_appends_only` that, when set to `True`, will only concatenate arrays that end in `"_append"`. Usually `utils.merge_dict()` will concatenate all arrays

### DIFF
--- a/fusesoc/capi2/inheritance.py
+++ b/fusesoc/capi2/inheritance.py
@@ -1,0 +1,39 @@
+# Copyright FuseSoC contributors
+# Licensed under the 2-Clause BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-2-Clause
+
+import copy
+import re
+
+from fusesoc import utils
+
+
+class Inheritance:
+    MERGE_OPERATOR = "<<__FUSESOC_MERGE_OVERLOAD__<<"
+
+    def yaml_merge_2_fusesoc_merge(capi):
+        """
+        Replace YAML merge key operator (<<) with FuseSoC merge operator
+        """
+        yaml_merge_pattern = (
+            r"((?:\n|{|\{\s*(?:[^{}]*\{[^{}]*\})*[^{}]*\},)\s*)<<(?=\s*:)"
+        )
+        while re.search(yaml_merge_pattern, capi):
+            capi = re.sub(yaml_merge_pattern, r"\1" + Inheritance.MERGE_OPERATOR, capi)
+        return capi
+
+    def elaborate_inheritance(capi):
+        if not isinstance(capi, dict):
+            return capi
+
+        for key, value in capi.items():
+            if isinstance(value, dict):
+                capi[key] = Inheritance.elaborate_inheritance(copy.deepcopy(value))
+
+        parent = capi.pop(Inheritance.MERGE_OPERATOR, {})
+        if isinstance(parent, dict):
+            capi = utils.merge_dict(parent, capi, concat_list_appends_only=True)
+        else:
+            raise SyntaxError("Invalid use of inheritance operator")
+
+        return capi
--- a/fusesoc/utils.py
+++ b/fusesoc/utils.py
@@ -16,6 +16,8 @@ except ImportError:
     from yaml import SafeDumper as YamlDumper
     from yaml import SafeLoader as YamlLoader
 
+from fusesoc.capi2.inheritance import Inheritance
+
 logger = logging.getLogger(__name__)
 
 
@@ -159,15 +161,18 @@ def yaml_fread(filepath, resolve_env_vars=False, remove_preamble=False):
     with open(filepath) as f:
         if remove_preamble:
             f.readline()
-        return yaml_read(f, resolve_env_vars)
+        return yaml_read(f.read(), resolve_env_vars)
 
 
 def yaml_read(data, resolve_env_vars=False):
     try:
+        data = Inheritance.yaml_merge_2_fusesoc_merge(data)
+        capi_data = {}
         if resolve_env_vars:
-            return yaml.load(os.path.expandvars(data.read()), Loader=YamlLoader)
+            capi_data = yaml.load(os.path.expandvars(data), Loader=YamlLoader)
         else:
-            return yaml.load(data, Loader=YamlLoader)
+            capi_data = yaml.load(data, Loader=YamlLoader)
+        return Inheritance.elaborate_inheritance(capi_data)
     except (yaml.parser.ParserError, yaml.scanner.ScannerError) as e:
         raise SyntaxError(str(e))
 
@@ -176,11 +181,13 @@ def yaml_dump(data):
     return yaml.dump(data)
 
 
-def merge_dict(d1, d2):
+def merge_dict(d1, d2, concat_list_appends_only=False):
     for key, value in d2.items():
         if isinstance(value, dict):
             d1[key] = merge_dict(d1.get(key, {}), value)
-        elif isinstance(value, list):
+        elif isinstance(value, list) and (
+            not concat_list_appends_only or key.endswith("_append")
+        ):
             d1[key] = d1.get(key, []) + value
         else:
             d1[key] = value
--- a/tests/capi2_cores/parser/inheritance.core
+++ b/tests/capi2_cores/parser/inheritance.core
@@ -1,0 +1,59 @@
+CAPI=2:
+# Copyright FuseSoC contributors
+# Licensed under the 2-Clause BSD License, see LICENSE for details.
+# SPDX-License-Identifier: BSD-2-Clause
+
+name: ::inheritance:0
+filesets:
+  fileset_a:
+    files:
+      - 1.txt
+      - 2.txt
+      - 3.txt
+  fileset_b:
+    files:
+      - 4.txt
+      - 5.txt
+      - 6.txt
+  fileset_c:
+    files:
+      - 7.txt
+      - 8.txt
+      - 9.txt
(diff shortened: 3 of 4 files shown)


olofk:
This is very clever! I have one concern though. Will this confuse other yaml readers? Just as an example, I would love to have a core file parser in JavaScript to embed in websites, but then I would need to implement the same override for that yaml parsing library, right?

sifferman:
The replacement of `<<` keys should be fine with other YAML readers. But yes, you would have to implement the `yaml_merge_2_fusesoc_merge()` and `elaborate_inheritance()` functions. The only weirdness I would expect between versions would be how references are handled. To get it to work with, PyYAML, I needed to run a bunch of `deep_copy`s (though there is probably a better solution)

Also, I think this PR will actually *improve* compatibility with other YAML readers considering that the merge operator has been depreciated in YAML 1.2 ([ref](https://stackoverflow.com/questions/64224811/yaml-merge-multiply-anchors-into-the-same-map)).

olofk:
Ouch... you're right. I had no idea. Hmm... knowing this, I wonder if we should do this in two stages.
1. Apply your patch, since that's probably more intuitive
2. Eventually come up with some explicit syntax to make reuse easier and at that point deprecate and warn against the use of << to avoid confusing different yaml readers that might implement this differently

sifferman:
Sounds good! I can resolve the merge conflicts and squash

On how to add an explicit syntax:

In this patch, users can already use `<<__FUSESOC_MERGE_OVERLOAD__<<` in place of `<<`. Therefore, the `Inheritance.MERGE_OPERATOR` could simply be replaced with a user-friendly operator (maybe `$<` instead of `<<__FUSESOC_MERGE_OVERLOAD__<<`).

Though I don't really see a need to do this. Right now, if they use the old one (`<<`), it will be silently renamed to the new operator. But also, as long as `<<` is replaced before the YAML reader sees it, then there shouldn't be any issues with parsing/support. The only issue I see is that users may be confused why FuseSoC treats `<<` differently than how the YAML 1.1 LRM states

olofk:
Great. Let's start with this. After I wrote my last comment I remembered that I did a quick PoC on an `inherit` tag earlier this year after discussions with a customer that wanted a feature like this. Their use-case was that they had a lot of cores and wanted to reduce the amount of boilerplate by inheriting from a common template. Long-term I think we could be looking at something like that, which can cover both that use-case and the functionality that people want from the deprecated merge key feature.

sifferman:
Alright, I believe it's ready. Thanks for the help, @olofk!

sifferman:
Actually, I just realized `Inheritence.yaml_merge_2_fusesoc_merge()` is not as good as it should be yet. Let me improve it a bit

sifferman:
Alright, NOW it should be good

I added the following tests to make sure that only valid merge key operators are replaced with `Inheritance.MERGE_OPERATOR`

https://github.com/olofk/fusesoc/blob/636c1662f2c971193517cda4378e809e999495bc/tests/capi2_cores/parser/inheritance.core#L55-L59

I had to modify the regex in `yaml_merge_2_fusesoc_merge()`

https://github.com/olofk/fusesoc/blob/636c1662f2c971193517cda4378e809e999495bc/fusesoc/capi2/inheritance.py#L14-L23

olofk:
Fantastic! Looks perfect. Thanks for your contributions. Picked and pushed!

PR #649, opened 2023-10-06
Add parameter indicating if files are exported

paul-gatewood:
This parameter is used for a PR to [enable the --no-export flag for Libero](https://github.com/olofk/edalize/pull/398). Libero's `import_file` command always copies the file into the local project directory, so the project script needs a way to tell if it should link instead using `create_link`.


olofk:
Will this still be needed with the latest updates to the Libero PR?

paul-gatewood:
No, with the simplification to always use ‘create_links’, this is no longer needed. Thanks again!

PR #658, opened 2023-12-26
"tool list' CLI argument now functional

phillipe87:
The "tool list" command line argument was returning nothing when invoked.

After quick inspection, I noticed an import statement was missing and was able to make the very minor fix.

### DIFF
--- a/fusesoc/main.py
+++ b/fusesoc/main.py
@@ -200,7 +200,7 @@ def list_cores(fs, args):
 
 
 def list_tools(fs, args):
-    from edalize.edatool import walk_tool_packages
+    from edalize.edatool import get_edatool, walk_tool_packages
 
     _tp = list(walk_tool_packages())
     maxlen = max(map(len, _tp))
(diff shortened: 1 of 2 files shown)


olofk:
Thanks! I completely forgot that command even existed. Come to think of it, we should probably add a similar command for the new Flow API as well.

phillipe87:
Which new Flow API? I am still familiarizing myself with the codebase.

olofk:
The Flow API is technically in Edalize (some info [here](https://edalize.readthedocs.io/en/latest/edam/api.html#flow-options) and on [my blog](https://blog.award-winning.me/2022/01/edalize-030.html) ), but we can select whether to use the old Tool API or the new Flow API from FuseSoC using the `flow` key in a target of a core file. If the `flow` key is present, it will use the Flow API, otherwise it will fall back to the old Tool API.

I still have on my todo list to better document how this all fits together, but at least there are some pointers above.

PR #659, opened 2023-12-29
Use pydantic for corefile validation

liambeguin:
Hi @olofk, 

I was looking at the core file parsing for other reasons and thought I'd put together a "demo" with Pydantic that I've been using in other projects. It's a library for data validation and parsing. It enforces type hints at runtime, and provides errors when data is invalid. It's a nice way to centralize all input processing.

It might be a big change so I understand if you're not really up for it, but still I thought I'd put something together.

```python
>>> import yaml
>>> from fusesoc.capi2 import model
>>>
>>> infile = 'tests/cores/jtag_tap/jtag_tap-1.13.core'
>>> with open(infile, 'r') as f:
...     data = yaml.safe_load(f)
... 
>>> try:
...     corefile = model.CAPI2(**data)
... except Exception as e:
...     print(e)
... else:
...     print(corefile)
... 
CAPI2=None name='::jtag_tap:1.13' description='JTAG Tap' provider=ProviderGithub(name='github', patches=[], cachable=True, user='olofk', repo='jtag', version='v1.13') filesets={'verilog_src_files': Fileset(file_type='verilogSource', logical_name=None, tags=None, files=['tap/rtl/verilog/tap_top.v'])}
>>> 
>>> 
>>> del data['CAPI=2']
>>> del data['privider']['user']
>>> 
>>> try:
...     corefile = model.CAPI2(**data)
... except Exception as e:
...     print(e)
... else:
...     print(corefile) 
... 
2 validation errors for CAPI2
CAPI=2
  Field required [type=missing, input_value={'description': 'JTAG Tap...['verilog_src_files']}}}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.5/v/missing
provider.github.user
  Field required [type=missing, input_value={'name': 'github', 'patch...ag', 'version': 'v1.13'}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.5/v/missing
```

There's also a sphinx plugin which looks something like this:

image

Let me know what you think and I can keep updating this.


olofk:
The idea of using jsonschema to validate core file is because I want it to be language-agnostic, so that someone implementing a core file parser in e.g. JavaScript or Rust can use the same schema. So for that reason I would vote against rewriting the schema in pydantic.

liambeguin:
> The idea of using jsonschema to validate core file is because I want it to be language-agnostic, so that someone implementing a core file parser in e.g. JavaScript or Rust can use the same schema. So for that reason I would vote against rewriting the schema in pydantic.

I understand, that makes sense even though this would simplify the existing code.

If it makes a difference, one can dump the the schema pretty easily from a BaseModel with something like this:

```python
import json # works with YAML too
from fusesoc.capi2 import model

schema = model.CAPI2.model_json_schema()
print(json.dumps(schema))
```

```console
$ python ./example.py | jq .
{                                                                             
  "$defs": {                                                                  
    "Files": {                                                                
      "additionalProperties": true,                                           
      "properties": {                                                         
        "is_include_file": {                                                  
          "anyOf": [                                                          
            {                                                                 
              "type": "boolean"                                               
            },                                                                
            {                                                                 
              "type": "null"                                                  
            }                                                                 
          ],                                                                  
          "default": null,                                                    
          "description": "\n        Treats file as an include file when true",
          "title": "Is Include File"                                          
        },                                                                    
        "include_path": {                                                     
          "anyOf": [                                                          
...
```

olofk:
That's good to know, but I think we'll keep it as it is. Long-term goal is to move the json schema to an external repository, and in that case I think it's better to have that as the source definition.

So, thanks for bringing up the proposal, but I'm closing this now.

PR #660, opened 2023-12-31
Update json_schema.py

hackathlete:
Several superfluous commas removed


olofk:
Thanks! Picked and pushed

PR #661, opened 2023-12-31
Modernize python packaging with pyproject.toml and build-backend

djcopley:
Adding a pyproject.toml with a build backend is the essential first step in modernizing a Python project, as it sets the foundation for utilizing PEP 517 and PEP 518 standards, enabling improved packaging, dependency management, and compatibility with modern build tools, as detailed in the packaging guide: https://packaging.python.org/en/latest/guides/modernize-setup-py-project/#where-to-start.

This update facilitates building and installation using the following recommended commands:

Deprecated | Recommendation
-- | --
python setup.py install | python -m pip install .
python setup.py develop | python -m pip install --editable .
python setup.py sdist | python -m build
python setup.py bdist_wheel | python -m build

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,3 @@
+[build-system]
+requires = ["setuptools"]
+build-backend = "setuptools.build_meta"


olofk:
Thanks! I've been meaning to look into modern Python packaging best practices, but never found the time. You say this is the first step. Is the goal to eventually get rid of setup.py or will it live side-by-side with the toml file in the future?

djcopley:
Great question! Either option works, really. The new packaging standards aim to separate the front end from the backend in building. This change effectively achieves that goal. Currently, using setuptools as a backend with a `setup.py` is not deprecated. The only notable drawback, in my opinion, is that people can still execute the command `python setup.py COMMAND`, which is deprecated.

That being said, fusesoc's setup.py is pretty straightforward and could be replaced entirely with a `pyproject.toml` super easily. I think it would probably be worthwhile at some point in the near future, given that this seems to be the way that all python packaging is going.

olofk:
Sounds like a nice simplification if we can get rid of setup.py completely in this case. Only thing to be aware of is to make sure it still works with older Python versions down to 3.6.

I'm pulling this in now and we can do the rest as follow-up PR. Thank you for your contributions!

PR #664, opened 2024-01-12
Replace setup.py with pyproject.toml

djcopley:
Follow on MR to #661


djcopley:
It seems like this isn't going to work well with python3.6. Closing for now.

olofk:
That's a shame. Let's revisit in the future when we have raised the minimum version. Thanks for looking into it

PR #665, opened 2024-01-15
Add release workflow for PyPI publish automation

djcopley:
Not sure if you're interested in this piece of automation, but this workflow will automatically build and publish a FuseSoC release to PyPI when you create a GitHub release.

I configured it to use OIDC for authentication and to run in the production environment. This means that you can set up special rules, such as mandatory review etc., before it can execute.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,39 @@
+name: Release
+on:
+  release:
+    types: [ published ]
+permissions:
+  contents: read
+
+jobs:
+  release-build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.x"
+      - name: Build release distributions
+        run: |
+          pip install build
+          python -m build
+      - name: upload windows dists
+        uses: actions/upload-artifact@v4
+        with:
+          name: release-dists
+          path: dist/
+  pypi-publish:
+    runs-on: ubuntu-latest
+    needs:
+      - release-build
+    environment: release
+    permissions:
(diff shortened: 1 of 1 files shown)


olofk:
This looks good. The current release process is pretty much undocumented, but I currently run

    python setup.py sdist
    twine upload dist/fusesoc-<version>.tar.gz

Is this doing the same thing more or less? Haven't use the `build` package before. And finally, I didn't get how authentication works. Currently I have a password in ~/.pypirc but I guess that's not used with this flow.

djcopley:
> Is this doing the same thing more or less?

It's doing basically the same. It uploads a wheel in addition. I think including the wheel is correct.

> And finally, I didn't get how authentication works.

There is a protocol called Open ID Connect (OIDC) that GitHub and PyPI support. To configure it you'll have to do a few things.

1. First go to the GitHub settings for this project and create an environment. This step is optional, but I highly recommend it for security. This is where you can set up all the approval and security rules. You can name it anything, however, it has to match what the workflow says. I named it '[release](https://github.com/olofk/fusesoc/pull/665/files#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R29)'.
2. Next go to your [PyPI publishing settings](https://pypi.org/manage/project/fusesoc/settings/publishing/) and go to the bottom where it says "Add a new publisher".  Fill in the repository info and save.

To use the workflow, just create a release in GitHub. If you set up approval rules, the actual publish to PyPI action won't happen until the required approvers approve the deployment.

Here is a github article on configuring OIDC:
https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-pypi

PR #668, opened 2024-01-20
coremanager: Pick preferred cores for virtual VLNVs

a-will:
Add "conflicts" constraints for each core implementing virtual VLNVs, so no two cores implementing the same virtual VLNV may be selected simultaneously.

This change should prevent more than a single implementing core from being selected as the implementation for a virtual VLNV, and when there is an explicit dependency in the tree, that one should be selected.

### DIFF
--- a/fusesoc/coremanager.py
+++ b/fusesoc/coremanager.py
@@ -113,6 +113,35 @@ class CoreDB:
     def solve(self, top_core, flags):
         return self._solve(top_core, flags)
 
+    def _get_conflict_map(self):
+        """Return a map of cores to their conflicts
+
+        Only one core that implements a virtual VLNV may be selected in a
+        dependency tree. For each core that implements a virtual VLNV, create a
+        set representing all other cores that implement one of the same virtual
+        VLNVs. In the resulting package definitions, these must get "conflicts"
+        constraints.
+        """
+        conflict_map = {}
+        virtual_map = {}
+        for core_data in self._cores.values():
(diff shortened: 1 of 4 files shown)


a-will:
The prior test case did not actually test the suggested specs outlined in https://github.com/olofk/fusesoc/issues/235
```
Now to use the bus, you need to explicitly choose which "provider" (implementation) of vendor:bus:config_default you want to include
```

This change still doesn't handle the case where no implementation is specified in the dependency tree (which probably should result in an error). In that case, fusesoc's dependency solver would pick effectively arbitrarily.

olofk:
I'm not super familiar with the code around the virtual / provides support, since that was written by other people, but I see the issue you are solving here and I trust it's doing the right thing. Just to be sure, is this one good to go, or do you want to fix the issue you mention in the last paragraph first?

a-will:
Offhand, I'm not sure how best to fix fusesoc's not reporting an error if no implementing core is part of the dependency tree. The modeling of a package repo with the SAT solver may not allow us to express the relationships we want.

So... this PR is at least an improvement that picks the desired core when it _is_ specified. Good to go, I'd say 😄

olofk:
FYI, I'm also unable to improve error messages. I have tried, but failed miserably at understanding the internals of the dependency manager I shoe-horned into FuseSoC. I really want to replace that with a custom one for several reasons, but until then we will have to live with it.

Picked and pushed! Thank you for your contributions

PR #672, opened 2024-02-21
Add better bash_completion with the argcomplete package

nicorum:
The current bash_completion script is very basic and the completion breaks after entering the first letter of a core name.
Since fusesoc uses the argparse package for argument parsing, we can use the argcomplete package for bash_completion.

Most of the completion is done automatically by argcomplete using the argparse parser.  So there is no need to replicate argument/subparser names in a separate bash_completion file.
I also implemented completion for cores, generators and tools for the specific arguments.

Completion for bash/zsh can be enabled either globally or on a per-command basis.
Global completion can be enabled with the _activate-global-python-argcomplete_  command. Additionally, the fusesoc binary needs the string PYTHON_ARGCOMPLETE_OK in the first 1024 bytes. I would have added this feature to the binary, but i can't find it in the repository. So it is not part of this pull request.
To enable completion for the fusesoc binary only, use the _eval "$(register-python-argcomplete fusesoc)"_ command.
Completion can also be used unofficially with other [shells](https://github.com/kislyuk/argcomplete/tree/develop/contrib) (e.g. powershell).

The disadvantage of this approach is that it adds the argcomplete package as a dependency. However, the argcomplete package seems to be well maintained and is part of the repository of many Linux distributions.

### DIFF
--- a/fusesoc/main.py
+++ b/fusesoc/main.py
@@ -1,4 +1,5 @@
 #!/usr/bin/env python
+# PYTHON_ARGCOMPLETE_OK
 # Copyright FuseSoC contributors
 # Licensed under the 2-Clause BSD License, see LICENSE for details.
 # SPDX-License-Identifier: BSD-2-Clause
@@ -10,6 +11,8 @@ import signal
 import sys
 import warnings
 from pathlib import Path
+
+import argcomplete
 
 try:
     from fusesoc.version import version as __version__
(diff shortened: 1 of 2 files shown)


nicorum:
I noticed that I had forgotten to use the pre-commit tool. The branch is now updated with the correctly formatted code.

nicorum:
Small update:
It seems that adding PYTHON_ARGCOMPLETE_OK to the binary generated by setuptools is not necessary. The argcomplete package automatically searches the script that is called by this binary.
I originally tested this under Ubuntu with python 3.8 where it didn't work, but under fedora with python 3.12 it works.
So this pull request should work with newer python versions as it is. For older versions there is always the option to add eval "$(register-python-argcomplete fusesoc)" to the bash_profile etc.

olofk:
Doing a long overdue scan of open PRs and I think it would be great to have this. Unfortunately we have moved from setup.py to pyproject.toml since this was submitted, but I think it should be a quite small change to fix that.

nicorum:
Updated to pyproject.toml

olofk:
Awesome! Thanks for this. Picked and pushed.

PR #684, opened 2024-04-04
Fix typos

nazar01:


olofk:
Thanks! A bit scary that one of the typos was in the code. I suspect no one actually tried using that feature :)

PR #695, opened 2024-06-24
Fix library update for git

rpatel3001:
Fetch instead of pull so tags work in sync-version, and do it first so we get updates before checking out the version.

### DIFF
--- a/fusesoc/provider/git.py
+++ b/fusesoc/provider/git.py
@@ -38,10 +38,10 @@ class Git(Provider):
 
     @staticmethod
     def update_library(library):
-        git_args = ["-C", library.location, "pull"]
+        git_args = ["-C", library.location, "fetch", "--all"]
         try:
-            Git._checkout_library_version(library)
             Launcher("git", git_args).run()
+            Git._checkout_library_version(library)
         except subprocess.CalledProcessError as e:
             raise RuntimeError(str(e))
 


rpatel3001:
Needs some work, this doesn't work for branches anymore

olofk:
Just let me know when you have something I should review. I remember that there were some tricks done to correctly get tags and branches.

rpatel3001:
not actively working on this but pasting a link for later

rpatel3001:
i don't know that anything fancy is really needed here, I reverted my previous change and just had it not print out the scary red error message if git returns non-zero. It still prints out the git output
```
You are not currently on a branch.
Please specify which branch you want to merge with.
See git-pull(1) for details.

    git pull <remote> <branch>
```
when sync-version is a tag or hash because the Launcher class is a pretty thin wrapper and you can't control that, but if a library is on a tag or commit maybe a user should just expect that. It might be desirable to have Launcher return the process output and re-raise the RuntimeError if that text isn't in the output, just in case there are other reasons git fails and you want the scary red message back?

olofk:
Was this solved by #724 ?

rpatel3001:
I think it mostly does. One thing I did in my original patch before I force pushed and overwrote it is swap the order of the fetch/pull and checkout. This might be required if eg the sync version is a hash or tag that's new in the remote repo so it doesn't exist in the local repo until after the fetch occurs.

Issue #616, opened 2023-03-27
Generator-Cache removal and --no-export


Steinegger:
Files generated by a generator are removed after the setup phase. When using the --no-export flag the source files aren't copied to the respective build directory, hence, the project creation fails due to missing source files.
I also tried with, the `cache_type: input` and `cache_type: generator` setting to keep the fusesoc config in place [as mentioned in the docs here](https://fusesoc.readthedocs.io/en/latest/user/build_system/generators.html), but they would still be deleted.

This seems unintended behaviour, because deleting the files if the--no-export flag is present will almost always brick the build.
A not so obvious workaround seems to be to specify a copyto attribute for each file in the generated core, if this is the intended way of handling generators it should be noted in the docs.

Versions:
Python 3.10.0
Fusesoc 2.0
edalize 0.4.1

olofk:
Confirmed. The --no-export case slipped through our net when we were testing. Sorry about that. I have a potential fix [here](https://github.com/olofk/fusesoc/tree/no-export-generators-fix) and would love some additional testing to see that I got it right this time.

You can install a local development version of FuseSoC by cloning and checking out the branch with
`git clone https://github.com/olofk/fusesoc -b no-export-generators-fix`, cd into fusesoc and run `pip install --user -e .`

Steinegger:
Thanks for the quick fix. I removed my copyto statements and installed the dev-version. The fix creates a `generator_cache` directory in the `build/*/*/` folder and works perfectly with the no-export flag. Without --no-export the files are created int the `C:\users\<username>\.cache\fusesoc\generator_cache\` directory and copied over to the build directory afterwards.

Interestingly it also helps with issue #613, because the generated directory is moved to the build directory

olofk:
Looks like we can close this

PR #704, opened 2024-09-12
Fixed Generator Cache with --no-export

### LINKED ISSUE #673: Generator Cache doesn't work with --no-export
When using the --no-export argument the generator cache gets placed in the `build/<core>/<target>-<tool>/generator_cache `folder instead of the `.cache/fusesoc/generator_cache` folder. But this folder is removed at the start of each fusesoc run. Which is why the generator_cache does not work with the --no-export argument.

I can fix this problem for me by removing `self.gen_root` from the line that contains `self.gen_root or self.core.cache_root,` in the `Ttptttg` class. But I don't want to create a pull request with the possible fix. Because I don't understand what the exact purpose of using self.gen_root is. Is this behaviour intended?

nicorum:
Fixes #673
The issue was caused by 90500180872ed26c3bee2c41c50be5ff0b49c2b4
I also checked, that this fix does not break  #616 again

### DIFF
--- a/fusesoc/edalizer.py
+++ b/fusesoc/edalizer.py
@@ -152,10 +152,7 @@ class Edalizer:
                     core.direct_deps.append(str(gen_core.name))
                     gen_core.pos = _ttptttg.pos
                     self._resolved_or_generated_cores.append(gen_core)
-                    if self.export_root and not (
-                        _ttptttg.is_generator_cacheable()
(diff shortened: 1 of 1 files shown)


olofk:
Thanks! Pulling this in now. FYI, I'm just about to make some other changes in this area to better control where temporary core file generation is performed, but I hope not to break this again.

PR #706, opened 2024-09-20
Added --clean flag to fusesoc run command

nicorum:
Adds a clean flag that explicitly cleans (deletes) the build directory when using the flow API.


olofk:
Looks good. Thanks for that. Picked and pushed.

PR #711, opened 2024-12-18
Non deterministic virtual warning

blueluna:
Adding a warning message when there is a non-deterministic selection of a virtual core during setup.

### DIFF
--- a/fusesoc/coremanager.py
+++ b/fusesoc/coremanager.py
@@ -225,14 +225,36 @@ class CoreDB:
         except NoPackageFound as e:
             raise DependencyError(top_core.name)
 
+        virtual_selection = {}
         objdict = {}
         if len(transaction.operations) > 1:
             for op in transaction.operations:
+                package_name = self._package_name(op.package.core.name)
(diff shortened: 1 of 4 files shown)


olofk:
@a-will @HU90m Will this solve the non-determinism for virtual cores in OpenTitan? (cc @razer6)

a-will:
> @a-will @HU90m Will this solve the non-determinism for virtual cores in OpenTitan? (cc @Razer6)

It's a useful warning, at least! So build issues can be more readily explained in case someone forgot to add the implementing core to a core file.

If the user _did_ specify which virtual core should be used, then https://github.com/olofk/fusesoc/pull/668 ensures that it actually _would_ get used.

The last interesting bit might be not requiring the VLNVs to be placed in a core file (i.e. a mechanism to choose the cores from a command line option, adding to "requirements" alongside the tree generated by the system core). I just had to update about 200 core files to add implementing core specs for lint targets in OpenTitan, where the feature might have reduced that down to a single flow config file. :sweat_smile:

PR #712, opened 2024-12-19
Check if subversion is available before running SVN provider test

blueluna:


olofk:
Awesome! Thanks! Picked and pushed

PR #713, opened 2024-12-20
Always build non-cached generators in work_root

olofk:
Before this change, non-cached generators were only built in work_root if --no-export was enabled which caused race conditions when doing multiple parallel builds.

With this change we also don't remove the temporary generator work directory as that can be useful to have around and will be cleaned out on subsequent builds to the same work_root anyway.

### DIFF
--- a/fusesoc/edalizer.py
+++ b/fusesoc/edalizer.py
@@ -141,7 +141,6 @@ class Edalizer:
     def run_generators(self):
         """Run all generators"""
         self._resolved_or_generated_cores = []
-        self._generated_core_dirs_to_remove = []
         for core in self.cores:
             logger.debug("Running generators in " + str(core.name))
             core_flags = self._core_flags(core)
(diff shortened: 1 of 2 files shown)


a-will:
Looks good to me! If it isn't there already, I'd recommend adding a comment in docs somewhere stating that a `work_root` should not be shared across multiple simultaneous fusesoc processes.

olofk:
Great. I can add the clarification about work root, but I was assuming that all kinds of things already broke if you tried to use the same work root with different processes. Out of curiosity, is this something you are using?

a-will:
> Great. I can add the clarification about work root, but I was assuming that all kinds of things already broke if you tried to use the same work root with different processes. Out of curiosity, is this something you are using?

No, or at least not that I am aware of!

I was merely suggesting offering clarity to users about fusesoc's limits. Unlike bazel, for example, fusesoc does not prevent the user from attempting to share a work_root, even if might only happen by accident.

PR #723, opened 2025-01-17
Don't leave back stray library after tests

jfrimmel:
The new test `test_library_update_with_initialize` introduced in #720 did use the `fusesoc_libraries`-directory in the repository root and, in contrast to other tests in that module, did not delete it afterwards. This causes a leftover library in the repository root after running the testsuite. This is unpleasant and a bug in the test.

This commit fixes it by cloning into the temporary directory, which was already created for just this use-case.


olofk:
That makes sense. Thanks for the fix.

PR #726, opened 2025-01-24
Fix erroneous relative path resolution

ivanvig:
Currently FuseSoC has two problems related to relative path resolution, 

The more severe one occurs when the fusesoc.conf file is in the current directory, in that case the code 

[files_read = self._cp.read(config_files)](https://github.com/olofk/fusesoc/blob/b604f53daf3c28aedd196c5cc1335f3161280201/fusesoc/config.py#L40)

returns only the file name, then in the function [_resolve_path_from_cfg](https://github.com/olofk/fusesoc/blob/b604f53daf3c28aedd196c5cc1335f3161280201/fusesoc/config.py#L95-L107) when `os.path.dirname` is called, an empty string is returned and when joined to form the final path we get a broken result

A second issue that I found is when referencing to a relative path starting with a dot, for example `./build_root` that dot stays in the path producing a result like `/abs/path/to/./build_root` which although is a valid path, fails path comparisons, like the ones that are done for `ignored_dirs`

This PR includes tests for both cases and fixes them.

### DIFF
--- a/fusesoc/config.py
+++ b/fusesoc/config.py
@@ -103,8 +103,8 @@ class Config:
         if os.path.isabs(expanded):
             return expanded
         else:
-            cfg_file_dir = os.path.dirname(self._path)
-            return os.path.join(cfg_file_dir, expanded)
+            cfg_file_dir = os.path.dirname(os.path.abspath(self._path))
+            return os.path.normpath(os.path.join(cfg_file_dir, expanded))
 
     def _path_from_cfg(self, name):
         as_str = self._cp.get(Config.default_section, name, fallback=None)
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -92,6 +92,47 @@ def test_config_relative_path():
             assert getattr(conf, name) == os.path.join(abs_td, name)
 
 
+def test_config_relative_path_starts_with_dot():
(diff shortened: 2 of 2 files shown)


ivanvig:
@olofk could you take a look at this PR?

olofk:
This looks like a nice cleanup of the paths, but the tests fail on macos. Could you take a look at the logs and see if you can figure out what potentially needs to be fixed?

ivanvig:
It seems that realpath was adding a parent directory in macos which made the assertions fail, changed for normpath which does just what I wanted. I guess it should be fixed now, if it fails again I will get a macos docker image to run the tests.

olofk:
Tests are passing now, so let's hope it stays that way. Thank you for your contributions. Picked and pushed!

olofk:
Huh! Now it failed again? F-in inconstent Python. If you have some bandwidth to look at this, it would be highly appreciated.

ivanvig:
Yep, I will look into it, I enabled the CI in my fork so I will test it there. It seems to be a problem related to [how macos handles the /var directory](https://stackoverflow.com/a/75475425), kinda annoying but it must be fixed.

ivanvig:
Fixed in #731

PR #727, opened 2025-01-30
[virtual-vlnv] Add CLI option to add virtual providers

a-will:
Add --add-vlnv to enable adding user-requested VLNVs that are not under the "system" or top core's explicit dependency tree. The intention of this argument is to allow users to specify cores that provide the implementations of virtual VLNVs in these "side" trees. The CLI option allows users to provide these without requiring someone to write a core file with the explicit deps, so targets may be reused across multiple different configurations, and users don't need to have duplicate core files that only differ by virtual providers.


a-will:
@olofk Here's a potential implementation of that command-line option I was talking about (for providing more dynamic insertion of deps to resolve virtual core references).

HU90m:
https://github.com/lowRISC/opentitan/pull/23555#issuecomment-2626811170

fusesoc/coremanager.py
```diff
@@ -249,11 +255,12 @@ def eq_vln(this, that):
                 ]
         # Print a warning for all virtual selections that has no concrete requirement selection
         for virtual in virtual_selection.values():
-            logger.warning(
-                "Non-deterministic selection of virtual core {} selected {}".format(
-                    virtual[1], virtual[0]
+            if virtual[0] not in requested_cores:
```
a-will, line 258: This matching process appears to be too strict. A request for a VLNV without the version specified doesn't match against a VLNV with the version included.

a-will:
> As mentioned in #728 I think we should sync up the virtual handling for consistency reasons. In the end I would like to see support for virtual->concrete mapping in a config file, on the command-line and in the target sections.
> 
> I'm also a bit sceptical about calling this add-vlnv and allowing to throw in cores that aren't necessarily in the dependency tree. From my understanding, the purpose of this functionality is not really to add extra VLNVs, but to guide FuseSoC towards using a particular concrete implementation of a virtual. I'm a bit worried about people abusing this to actually bring in extra cores that are completely outside of the dependency tree and that this could potentially cause issues further along.

This is a fair concern. We'll maybe need to think about how to  get all the power we want from --add-vlnv without the downsides.

> So I would suggest something like `--virtual some:virtual:vlnv=a:concrete:implementation:0.2` and then we might be able to use the business logic of #728 to load this mapping from the command-line instead of from a config file.

The downside to this exact approach is that it requires explicitly handling _every_ virtual VLNV individually. There is no straightforward way to collect a group of them into a core representing a library of them and succinctly adding the library of providers on the command-line.

Perhaps this is where a lockfile fragment or something could come in, though. Maybe what makes --add-vlnv and virtual provider libraries get into trouble is representing the mapping as a fusesoc core.

If we are to use a different kind of file to describe such a virtual provider library, then we'd need the ability to specify multiple mapping files to use (multiple libraries), and the paths to the mapping files would be a necessary argument, I think.

olofk:
> The downside to this exact approach is that it requires explicitly handling _every_ virtual VLNV individually. There is no straightforward way to collect a group of them into a core representing a library of them and succinctly adding the library of providers on the command-line.
> 

Ok, so for my understand, was your idea is to bring in an extra VLNV that depends on all the concrete implementations to have them all set at once? If so, wouldn't this also bring in unused dependencies as well?

> Perhaps this is where a lockfile fragment or something could come in, though. Maybe what makes --add-vlnv and virtual provider libraries get into trouble is representing the mapping as a fusesoc core (i.e. a _new_ VLNV for the virtual provider library).
> 
> If we are to use a different kind of file to describe such a virtual provider library, then we'd need the ability to specify multiple mapping files to use (multiple libraries), and the paths to the mapping files would be a necessary argument, I think.

I think this is the way forward. Let's extend #728 to allow loading a custom lockfile (or probably multiple, because, why not?) that you can use for virtual->concrete mapping.

andreaskurth:
I'm not sure lockfiles are the right place to implement the feature that is proposed in this PR. 

The use case that I think won't work well goes along the following lines: As a FuseSoC user, I want to select VLNV implementation providers when I invoke FuseSoC, without having to modify the code base (including `.core` or `.lock` files). As a concrete example, consider the case in which I have a code base and would like to use one set of VLNV implementation providers when I build an FPGA bitstream, another set when I run top-level simulations with tech-specific models, and a third set when I run simulations with generic models.

If this was implemented through lockfiles, I'd have to select different combinations of lockfiles for each of the runs, right? This could work but would be less convenient than the command line argument proposed in this PR.

To select implementation providers that are not part of the code base (e.g., proprietary implementations of modules that have a generic open-source implementation), I'd even have to modify parts of the lockfiles, right? In contrast, with the command line argument proposed in this PR, I'd just have to make sure the local implementation providers can be found by FuseSoC and tell it to prefer them with a command line argument.

a-will:
> Ok, so for my understand, was your idea is to bring in an extra VLNV that depends on all the concrete implementations to have them all set at once? If so, wouldn't this also bring in unused dependencies as well?

That's right! Bringing in unused dependencies was a bad side effect of this approach, haha. For OT, it ended up not being _too_ terrible because we are only using virtual VLNVs for leaf nodes in the tree--This is all for targeting different technology libraries with common interfaces. 😄 

> I think this is the way forward. Let's extend #728 to allow loading a custom lockfile (or probably multiple, because, why not?) that you can use for virtual->concrete mapping.

Sounds good to me! The reason for multiple files is to allow reusing descriptions of libraries of virtual->concrete maps.

In OT, we have collections of virtual cores that represent the technology library (which we call "prim libraries"). Each file representing a set of virtual->concrete maps can be such a library. In the OT repo, we can check in these files, and the user can point a fusesoc run to the desired file that was manually written to use our virtual maps.

Having the ability to specify multiple files lets someone else do a similar concept for whatever virtual cores they come up with, and OT and their repo can coexist / be used in the same design.

I will say that using file paths as arguments for inputs seems kind of different for fusesoc, which tries to discover all the things and give the user the ability to reference things by "name." A different option would be to have virtual maps as a first-class concept with names (VLNVs again?), and the user could add these named maps instead of direct file paths.

I don't have strong opinions on one implementation vs. the other.

olofk:
@andreaskurth After I had understood the use-case better I agree that we need to be able to do this on the command-line. I do however think we will have the possibility to do that now. The original proposal from @a-will was to a) create a wrapper .core file that depended on all the concrete implementations for each specific primlib and b) use the proposed `--add-vlnv` to add that core to the dependency graph and thus force FuseSoC to use those impls for the requested proposals. I.e. one "config" file (the wrapper .core file) and one CLI arg (--add-vlnv).

The latest suggestion is instead to a) create .lock file (that only contains virtual->concrete mappings, no actual version locks) for each primlib and b) add a `--lockfile` CLI arg to point out the lockfile of interest. That would also be one "config" file (the lock file) and one CLI arg (--lockfile) but it would avoid the side-effects of pulling in unnecessary deps and the potential abuse of the dependency graph. By allowing multiple lock files, we can hopefully cater to more complex flows as well.

I do understand that this is not something that normally goes into lock files in other languages, but I still think this is extremely similar in scope (locking a specific implementation or locking a specific version) that I'd say it makes sense to have this as a part of the lock file functionality. @a-will pointed out that we don't normally point out file locations on the FuseSoC command-line, but we do have functions to set locations for cores-root, build-root, log-file, config file etc, so I think it's fine this way.

Now there are a couple other things to note here.
1. If we didn't have the requirement of allowing secret implementations of virtual cores I would had suggested using flags to control this instead to avoid the virtual handling in the first place.
2. If you know at the top-level what concrete implementations you need (e.g. for an FPGA implementation where you really only have one choice), you can make sure these gets used without lock files, just by adding them as dependencies in a fileset that gets pulled in by that target. I'm also open for adding a key to the target section to set virtuals directly. I believe that in most cases, the target is implicitly deciding which tech to use.

a-will:
> Now there are a couple other things to note here.
> 
> 1. If we didn't have the requirement of allowing secret implementations of virtual cores I would had suggested using flags to control this instead to avoid the virtual handling in the first place.
> 2. If you know at the top-level what concrete implementations you need (e.g. for an FPGA implementation where you really only have one choice), you can make sure these gets used without lock files, just by adding them as dependencies in a fileset that gets pulled in by that target. I'm also open for adding a key to the target section to set virtuals directly. I believe that in most cases, the target is implicitly deciding which tech to use.

For what it's worth, it's not only about secret implementations. Because of the common prim library interfaces, OT's fusesoc targets in one core file become retargetable! For example, at the block level, we can simulate a block against a pure behavioral implementation (our generic prim library) or against primitives designed for a specific technology.

Someone could come along and design a prim library based on the open source skywater PDK, for example, check that all the IPs simulate properly with the same tests we use for the generic prims, then use the validated IPs + prims in a much larger design to create a real chip. At least the IP block level would use exactly the same fusesoc targets, regardless of which prim library is chosen.

Moreover, the possible set of implementations isn't limited to what we can hard-code in the fusesoc core files upstream. The downstream implementations aren't necessarily secret; they simply are allowed to multiply without needing upstream to know they exist. Thus does fusesoc achieve more of the mission to create an ecosystem of reusable IPs. 😄

HU90m:
I wrote while I was thinking this problem through:

---

# The OpenTitan primitive problem

Prior reading:

- [How OpenTitan handles primitives](https://github.com/HU90m/fusesoc-examples/blob/main/prims_lib/doc/how-opentitan-handles-primitives.md)
- [The virtual cores approach to handling primitives][virtual cores approach]

[virtual cores approach]: https://github.com/HU90m/fusesoc-examples/tree/main/prims_lib#the-virtual-cores-approach
[virtual cores PR]: https://github.com/lowRISC/opentitan/pull/23555

There is a work in progress [virtual cores PR][] that moves primitives over to using virtual cores, but it currently has a problem that's not easily solvable without upstream FuseSoC changes.

### Invokable cores and setting concreate implementations

An invokable core is a core that is expected to be run directly. Some may call this a top-level core, but top-level has connotations that don't always apply to invokable cores.

Currently invoke a core with chosen concrete implementations, these implementations need to be available in within it's dependencies. In a small project, this makes sense; one can make the invokable core depend upon a core that has all the concrete cores needed in it's dependencies. In [fusesoc-examples][virtual cores approach],
`hugom:prims_generic:all` is an example of one of these.

fusesoc-examples uses FuseSoC's flags to allow concrete implementations to be swapped out using only the CLI.

``` yaml
filesets:
  files_top:
    depend:
      - "!not_prims_generic ? (lowrisc:prims_generic:all)"
      - "prims_specific ? (lowrisc:prims_specific:all)"
      - "prims_secret ? (lowrisc:prims_secret:all)"
```

``` sh
fusesoc run --flag not_prims_generic --flag prims_specific lowrisc:example:top
```

### OpenTitan's problem with concrete dependencies in core files

The OpenTitan project has many invokable cores beyond the different top-levels. A large number of these come from the lint targets for each of the hardware blocks, see [`hw/ip/csrng/csrng.core`][].

[`hw/ip/csrng/csrng.core`]: https://github.com/lowRISC/opentitan/blob/master/hw/ip/csrng/csrng.core

To set the desired concrete implementation of a primitive library for a lint run. One would have to add the following to every lint file:

``` yaml
  files_lint:
    depend:
      - lowrisc:prim_generic:all

# ...

  lint:
    <<: *default_target
    filesets_append:
      - files_lint
```

#### Swapping primitives

This really becomes a problem when an OpenTitan user working on a chip targeting a particular PDK wants to run lints using their primitives which target this PDK.
They'll have to edit every one of these files.

### An additional constraint on the solution

Currently the [virtual cores PR][] follows the approach in [fusesoc-examples][virtual cores approach] where a hardware block is able to depend on a virtual core
that includes all primitives available called `lowrisc:prim:all`.

This was done because it is currently allowed in OpenTitan, but it has lead to hardware blocks including a lot of dead code. Because of this, OpenTitan would like to depreciate `lowrisc:prim:all`.


## Solutions
### A simple solution: `--add-vlnv`

The simplest solution to this problem implement a `--add-vlnv` CLI argument which allows cores to be appended the invoked cores dependencies.

```sh
fusesoc --add-vlnv lowrisc:prim_generic:all lint lowrisc:ip:csrng
```

This means concrete implementations don't have to be added to each invokable core's file and swapping primitives becomes easy, *but flexibility of this argument introduces a big foot gun*.

### The substitution approach

Instead of `--add-vlnv` a more specialised `--virtual` argument could be introduced, which explicitly maps a virtual core to a concrete implementation that may or may not be within the invoked cores dependencies.

```sh
fusesoc --virtual lowrisc:prim:all=lowrisc:prim_generic:all lint lowrisc:ip:csrng
```

This works great when `lowrisc:prim:all` is used everywhere, but OpenTitan wants hardware blocks to have more granular primitive dependencies and is looking to depreciate `lowrisc:prim:all`. Therefore, many `--virtual` mapping arguments would need to be supplied, one for each primitive used.

What's would make this more ergonomic are pre-written groups mappings set in FuseSoC's configuration `fusesoc.conf`, which can then be used as a single `--virtual-mappings` argument. This does still require people to write a new set of mappings in `fusesoc.conf` when swapping primitives out for their own.


### The library approach

FuseSoC already has the notion of a grouping of core files: libraries. A library is a local file path or external repository URI containing core files. A user can declare an ordered list of libraries in a local or global `fusesoc.conf` file. FuseSoC looks through this list in order when searching for a core file. If two core with the same VLNV exist, the core encountered earlier in the list will be discarded. This allows one to override a library core file by adding a library with the same core later in the list, either by adding it to `fusesoc.conf` or pointing to it with `--cores-root`.


libraries-diagram


_An example of a `fusesoc.conf` file:_

``` ini
[library.fusesoc-cores]
sync-uri = https://github.com/fusesoc/fusesoc-cores
sync-type = git
```

If a concrete implementation for a virtual core doesn't exist in the invoked core's dependency tree, FuseSoC will search all cores for an implementation. So long as FuseSoC follows the same rules as when searching for cores, i.e. a concrete implementation in a later library takes precedence, this would be an elegant solution to the problem.

OpenTitan could have the following library list in `fusesoc.conf`:

``` ini
[library.opentitan-ip]
location = ./hw/ip

[library.opentitan-tops]
location = ./hw/top

[library.opentitan-prim-generic]
location = ./hw/prim/generic
```

When one wants to use their own primitives, they can just point to them with just the command line `--core-root`:

``` sh
fusesoc --cores-root=../prim_custom lint lowrisc:ip:csrng
```

For any virtual cores that aren't implemented using their own primitives library, the generic implementation will continue to be used.


#### Explicit dependencies still work

The OpenTitan `fusesoc.conf` would likely actually be:

``` ini
[library.opentitan-ip]
location = ./hw/ip

[library.opentitan-tops]
location = ./hw/top

[library.opentitan-prim-xilinx]
location = ./hw/prim/xilinx

[library.opentitan-prim-xilinx-ultrascale]
location = ./hw/prim/xilinx_ultrascale

[library.opentitan-prim-generic]
location = ./hw/prim/generic
```

This brings the FPGA primitives into FuseSoCs search so that cores targeting a specific FPGA, such as `lowrisc:systems:chip_earlgrey_cw310:0.1`, could explicitly ask for `prim_xilinx` the primitives.

#### Argument naming

A better name for `--cores-root` may be something along the lines of `--append-library` to better describe what it does. FuseSoC could also consider supporting multiple of this arguments for when one wants to add multiple libraries from the CLI.


### The new file type approach

A new file type or core feild could be introduced to hold virtual mappings. A core field could look something like:

``` yaml
CAPI2:

name: "lowrisc:prim_generic::all:0.1"
mapping:
  "lowrisc:prim:sram_1p": "lowrisc:prim_generic:sram_1p"
  "lowrisc:prim:sram_2p": "lowrisc:prim_generic:sram_2p"
```

This core's vlnv could then be provided as a CLI argument.

``` sh
fusesoc --virtal-mapping=lowrisc:prim_generic:all lint lowrisc:ip:csrng
```

When someone want to add their own primitives, they can provide a mapping file with these primitives. This means they can swap out primitives without having to modify existing files.

HU90m:
FuseSoC [has settled](https://github.com/olofk/fusesoc/pull/728) on implementing a core-based lockfile (as oppose to library-based). It will also accept partial lock files. This means that the lock file can also be used similarly to a mapping file: the last solution outlined in https://github.com/olofk/fusesoc/pull/727#issuecomment-2647695340 .

People implementing a new set of primitives for OpenTitan will be able to make a partial lock file with their concrete implementations. They can then provide this lockfile to FuseSoC as an argument, `--lockfile=custom_prims.lock`, and these implementations will be used over any other implementations:

```yaml
lockfile_version: 1
fusesoc_version: 2.4.2
cores:
- "lowrisc:prim_generic:sram_1p"
- "lowrisc:prim_generic:sram_2p"
# ...
```

HU90m:
I should add partial lock files aren't nearly as ideal as mapping files. Because, instead of a clear mapping of virtual core to concrete implementation, partial lock files just force the use of a particular core.

If a set of concrete implementations are given in a partial lock file and there are no other implementations in the dependency tree, these concrete implementations will be used. *It's equivalent to using the `--add-vlnv` argument proposed in this PR.* However, if there is another set of implementations that exist in the dependency tree, it's ambiguous as to which will be selected.

So invokable OpenTitan cores will need to avoid adding concrete implementations in their dependency trees, so that any added in the lock file will be deterministically selected. This means invokable core will not be usable without specifying a lock file.

The library and mapping file approaches both allow primitives to be overridden.

@a-will are there any other points, clarifications or concerns that I've missed?
@olofk is my understanding of the proposed partial lockfiles correct?

olofk:
I honestly lost track of the progress here, but I'll close it and see if someone screams :)

PR #731, opened 2025-02-14
Fix canonical path vs symlink comparison

ivanvig:
This PR fixes the macOS tests, which were broken by pull request #726 . The issue occurs because on macOS, **/var** and **/tmp** are symbolic links to **/private/var** and **/private/tmp**. When calling `abspath` on a relative path or filename, we get the canonical (real) path. However, the `TemporaryDirectory()` function used in the tests returns a symbolic link that is already an absolute path (starting with **/var**), so calling `abspath` on it does nothing.

I also replaced the `abspath` call when processing the ignored directories with `realpath` to avoid a mismatch between a symlink and the canonical path of the same directory.

Everything seems to be working now, but I'm still a bit worried that there might be another comparison between symlink paths and real paths elsewhere in the code, since `abspath` is used in multiple places. A more comprehensive solution might be to replace all `abspath` calls with `realpath`, but I didn't want to make such an intrusive change to the codebase.

### DIFF
--- a/fusesoc/config.py
+++ b/fusesoc/config.py
@@ -103,7 +103,7 @@ class Config:
         if os.path.isabs(expanded):
             return expanded
         else:
-            cfg_file_dir = os.path.dirname(os.path.abspath(self._path))
+            cfg_file_dir = os.path.dirname(os.path.realpath(self._path))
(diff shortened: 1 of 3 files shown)


olofk:
Percecto. Gracias por eso!

PR #735, opened 2025-02-25
Add ruff lints

HU90m:
I noticed a few problems in the code base that would have been picked up on by a linter.

I've added the ruff linter to `pre-commit` and addressed some of the failures with simpler fixes.

Do we think using ruff for linting is a good idea?


olofk:
I don't have any opinions on what linter to use, but it looks like it picked up some nice clean-ups at least, so I'm happy to give it a go. Want to pull some other things in first though to avoid merge conflicts.

erikwa:
Sorry, did not notice your PR before I had almost finished with my changes :)

PR #739, opened 2025-03-20
Introduce the mappings feature

HU90m:
Introduce the mappings feature discussed in https://github.com/olofk/fusesoc/pull/727. *See https://github.com/olofk/fusesoc/pull/727#issuecomment-2647695340 for context.*

Description of the feature from the documentation:
> Mappings allow a user of a core to substitute dependencies of the core with other cores without having to edit any of the core or it's dependencies files. An example use case is making use of an existing core but substituting one of it's dependencies with a version that some desired changes (e.g. bug fixes).

I've restricted the flexibility of mappings in the interest of keeping this patch set smaller:
- They cannot be recursive.
- They cannot include versions.
- Cores conflicting with substitutes aren't removed from the dependency tree.
  - For OpenTitan's use case, a flag can be used to disable the default primitives so there are no conflicts.

These restrictions can be lifted in the future commits, if deemed useful.

The commits are all self contained: lints and tests pass for every commit, so they don't need to be squashed. However, re-ordering may make sense.

### DIFF
--- a/doc/source/user/build_system/index.rst
+++ b/doc/source/user/build_system/index.rst
@@ -38,5 +38,7 @@ A full reference documentation on the CAPI2 core file format can be found in the
    filters.rst
    flags.rst
    generators.rst
+   virtual_cores.rst
+   mappings.rst
    hooks.rst
    vpi.rst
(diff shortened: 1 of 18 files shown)


olofk:
There's one potential feature I could be interested in, and that's the possibility to set single mapping directly on the command line, e.g. `--map some:virtual:core some:specific:core` but it's more of a convenience than anything strictly needed.

PR #741, opened 2025-05-27
Export generator files successfully

paul-gatewood:
Since items the src_files list start with './' for files from a generated core (or at least for the core I'm experimenting with), the files should be normalized before comparison. Otherwise, all of the generated files will be removed, and no export takes place.


olofk:
Excellent! I have been bitten by this bug too but never found the time to investigate. Picked and pushed! Thanks.

PR #744, opened 2025-06-24
Add cli option to publish to core db.

anders-ahlberg:
This depends on changes in pending pull request #743 (signing and verifying).

### DIFF
--- a/fusesoc/main.py
+++ b/fusesoc/main.py
@@ -18,7 +18,7 @@ from pathlib import Path
 import argcomplete
 import requests
 
-from fusesoc import signature
+from fusesoc import publish, signature
 
 try:
     from fusesoc.version import version as __version__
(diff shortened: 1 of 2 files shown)


olofk:
We need some documentation as well.

fusesoc/main.py
```diff
@@ -298,6 +301,127 @@ def core_sign(fs, args):
...
+        guess["name"], user, repo, version
+    )
+    guess["found"] = True
+    return guess
+
+
+def core_publish(fs, args):
+    core = _get_core(fs, args.core)
+    uri = fs.config.publish_uri
+    pem = fs.config.publish_uri_pem
```
olofk, line 351: Why do we need this?

fusesoc/main.py
```diff
@@ -298,6 +301,127 @@ def core_sign(fs, args):
...
+        if c != "yes":
+            print("Aborted.")
+            return False
+
+    target = uri + "/v1/publish/"
+    logger.debug("POST to " + target)
+    res = requests.post(target, files=body, allow_redirects=True, verify=pem)
+    if not res.ok:
+        print("Request returned http result", res.status_code, res.reason)
+        err = json.loads(res.content)
```
olofk, line 417: Need to catch more errors here. If the server responds 404 it looks like there isn't a valid json reply at all
```
Request returned http result 404 Not Found
  File "/home/olof/projects/serv/bin/fusesoc", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/olof/projects/fusesoc/fusesoc/fusesoc/main.py", line 950, in main
    fusesoc(args)
  File "/home/olof/projects/fusesoc/fusesoc/fusesoc/main.py", line 940, in fusesoc
    args.func(fs, args)
  File "/home/olof/projects/fusesoc/fusesoc/fusesoc/main.py", line 417, in core_publish
    err = json.loads(res.content)
          ^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/json/__init__.py", line 346, in loads
    return _default_decoder.decode(s)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/json/decoder.py", line 355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 2 column 1 (char 1)
```

fusesoc/main.py
```diff
@@ -298,6 +301,127 @@ def core_sign(fs, args):
...
+        print(json.dumps(err, indent=4))
```
olofk, line 418: Let's try to handle known errors, e.g. validation errors, in a more user-friendly way. Instead of
```
Request returned http result 400 Bad Request
{
    "non_field_errors": [
        "Validation error in core::: 'description' is a required property"
    ]
}
```
we could look for `non_field_errors` and write something like  `Validation error: 'description' is a required property`

fusesoc/main.py
```diff
@@ -298,6 +301,127 @@ def core_sign(fs, args):
     print(f"{sigfile} created")
 
 
+def guess_provider():
```
olofk, line 304: Since there is quite a lot of new code, I would prefer to have this in a separate `publish.py`. `main.py` is already way too large for its own good.

fusesoc/main.py
```diff
@@ -298,6 +301,127 @@ def core_sign(fs, args):
...
+    guess = {"found": False}
+    cmd = ["git", "remote", "-v"]
+    res = subprocess.run(cmd, capture_output=True).stdout.decode("utf-8").strip()
```
olofk, line 307: Let's run this from the root directory of the core instead. We can pick up the directory from `core.core_root` in `core_publish` and send it to `guess_provider` as an argument. The reason is that most FuseSoC users have a workspace which is separate from the directories where the cores are.

fusesoc/main.py
```diff
@@ -298,6 +301,127 @@ def core_sign(fs, args):
...
+    if os.path.exists(sigfile):
+        print("and signature file: " + sigfile)
+        fob_sig = open(sigfile, "rb")
+        body["signature_file"] = fob_sig
+    else:
+        print("(without signature file)")
+        sf_data = None
+    if pem:
+        print("with certificate from: " + pem)
+    print("to api at: " + uri)
```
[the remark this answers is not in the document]
anders-ahlberg: Do we want a hardcoded default uri?  If so, what should it point to?

fusesoc/main.py
anders-ahlberg: Do you mean specifically the pem option?  It was needed during development since that server was self signed, and why not keep the option?

fusesoc/main.py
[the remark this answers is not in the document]
anlu85: If you like to add a default, pointing to the "official" FOSSI Foundation server instance might be a good option:
[https://fusesoc.fossi-foundation.net/](https://fusesoc.fossi-foundation.net/)

PR #748, opened 2025-08-13
Signatures

anders-ahlberg:


fusesoc/main.py
```diff
@@ -187,6 +189,14 @@ def library_list(fs, args):
 
 def list_cores(fs, args):
     cores = fs.get_cores()
+    if not fs.config.ssh_trustfile:
+        logger.info(
+            "NOTE: No trustfile configured (ssh-trustfile in fusesoc.conf), signatures will not be checked."
+        )
+    elif not os.path.isfile(fs.config.ssh_trustfile):
+        logger.info(
```
olofk, line 197: Same as above

fusesoc/main.py
```diff
@@ -268,7 +280,28 @@ def gen_show(fs, args):
...
+    sig = signature.sign(core.core_file, args.keyfile, None)
+    file = open(sigfile, "w")
+    file.write(sig)
+    file.close()
+    print(f"{sigfile} created")
+
+
+def core_verify(fs, args):
+    core = _get_core(fs, args.core)
+    res = signature.verify(core.core_file, args.trustfile, args.sigfile)
```
olofk, line 301: Drop args.trustfile and use `fs.config.ssh_trustfile` instead (more on that below). I think we can also be strict about requiring sigfile to be called `core_file+".sig"` so that we can drop args.sigfile as well. Hmm... come to think of it, do we even need this function when we have `fusesoc core show`?

fusesoc/main.py
```diff
@@ -501,6 +534,25 @@ def get_parser():
...
+    parser_core_sign = core_subparsers.add_parser(
+        "sign", help="Create user signature for a core"
+    )
+    parser_core_sign.add_argument(
+        "core", help="Name of the core to sign"
+    ).completer = CoreCompleter()
+    parser_core_sign.add_argument("keyfile", help="File containing ssh private key")
+    parser_core_sign.set_defaults(func=core_sign)
+
+    parser_core_verify = core_subparsers.add_parser(
```
olofk, line 546: Let's decide whether we want to keep the verify subcommand at all. I think we can just use `fusesoc core show` for this. With that said, I think we need to be more coherent wrt the trust file and make it possible to specify it both in fusesoc.conf and on the command-line for commands that need it. I think we should add a `--ssh-trustfile` argument (maybe with a different name) to the global parser. That way we can hook it up to the config machinery and set it on the command-line with a fallback to fusesoc.conf

fusesoc/signature.py
```diff
@@ -0,0 +1,173 @@
...
+import tempfile
+
+import yaml
+
+from fusesoc import utils
+
+logger = logging.getLogger(__name__)
+
+
+def read_core(data_file_name):
```
olofk, line 21: Let's put this function into the core object instead (and call it someting like get_data_to_sign, but preferably catchier :) ). That way the core itself is responsible for presenting the data to be signed, which gives us a bit more freedom to change things in future CAPI versions.

fusesoc/signature.py
```diff
@@ -0,0 +1,173 @@
...
+        # Core file is single document, no built in signature.  We
+        # sign everything after the header line.
+        core_canonical = core_raw[7:].strip()
+    else:
+        # Core file is not a valid CAPI=2 document.
+        raise RuntimeError("Signature file is not a valid CAPI=2 document.")
+    return core_canonical
+
+
+def sign(data_file_name, key_file_name, old_sig_file):
```
olofk, line 39: I still want the core object to be passed here instead. The actual data file is to me a private detail of the core. We get the name directly by `core.name` and avoid having to read the .core file as a yaml file (which might change in future CAPI versions). The actual data to sign should be fetched from the core through an api call, e.g. `core_canonical = core.get_data_to_sign()`

fusesoc/main.py
anders-ahlberg: Fixed in update.

fusesoc/main.py
anders-ahlberg: Function removed.

fusesoc/main.py
anders-ahlberg: Function removed.

fusesoc/capi2/core.py
```diff
@@ -638,3 +640,48 @@ def get_description(self):
     @property
     def mapping(self) -> Optional[Mapping[str, str]]:
         return MappingProxyType(self._coredata.get("mapping", {}))
+
+    def sig_status_long(self, trustfile):
+        if not trustfile:
```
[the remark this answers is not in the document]
anders-ahlberg: Fixed in update.

fusesoc/signature.py
```diff
@@ -0,0 +1,170 @@
...
+
+    if shutil.which("ssh-keygen") == None:
+        raise RuntimeError("ssh-keygen not found in $PATH")
+
+    cmd = ["ssh-keygen", "-Y", "sign", "-f", key_file_name, "-n", "file"]
+    logger.info("Run command: " + " ".join(cmd))
+    res = subprocess.run(cmd, input=core_canonical, capture_output=True)
+    sig_data = {
+        "coresig": {
+            "name": str(core.get_name()),
```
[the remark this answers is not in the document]
anders-ahlberg: Fixed.

fusesoc/signature.py
```diff
@@ -0,0 +1,170 @@
...
+    logger.debug("with trustfile:    " + trust_file_name)
+    core_canonical = core_obj.signed_data()
+    sig_data = utils.yaml_fread(sig_file_name)
+    if not "coresig" in sig_data:
+        raise RuntimeError("Signature file missing coresig member.")
+    if not isinstance(sig_data["coresig"], dict):
+        raise RuntimeError("coresig object in signature is not an object.")
+    if not "name" in sig_data["coresig"]:
+        raise RuntimeError("Signature file missing name member in coresig object.")
+    if sig_data["coresig"]["name"] != str(core_obj.get_name()):
```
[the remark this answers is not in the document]
anders-ahlberg: Fixed.

doc/source/user/build_system/core_files.rst
```diff
@@ -270,3 +270,27 @@ As always with inheritance the interesting questions are around overriding behav
   To do so, it specifies the special ``filesets_append`` key with a partial list.
   When evaluating the core file, FuseSoC appends the contents of ``sim.fileset_append`` at the end of ``default.fileset`` to form a list with two items: ``rtl``, and ``tb``.
   The same behavior works for all lists in core files.
+
```
olofk, line 273: This looks great. Just need to add some more text about how to sign cores

doc/source/user/build_system/core_files.rst
```diff
@@ -270,3 +270,29 @@ As always with inheritance the interesting questions are around overriding behav
...
+   user2@example.com ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIP6w+b8Xue0Jo/Hskd9B+Ttr6g9rq99qbZV/01q+FgED
+
+The location of this file can be set in the ``[main]`` section of the ``fusesoc.conf`` file:
+
+.. code-block:: INI
+
+   [main]
+   ssh-trustfile = /home/anders/FuseSoCdb/git/fusesoc/tests/signature_files/trustfiles/trustfile
+
+This setting can also be overridden using the command line option ``--ssh-trust-file``.
```
olofk, line 294: s/trust-file/trustfile

olofk:
Excellent! Looks perfectly fine now. Squashed and merged. Thanks for this

PR #750, opened 2025-09-04
Adds license field to CAPI2 schema

anlu85:
Adds support for specifying license information in the CAPI2
schema. The license can be provided as an SPDX identifier string
or as a custom license object with name and text attributes.

This allows for better documentation and identification of the
licenses used in FuseSoC cores.

### DIFF
--- a/fusesoc/capi2/json_schema.py
+++ b/fusesoc/capi2/json_schema.py
@@ -9,6 +9,24 @@ capi2_schema = """
       "description": "Short description of core",
       "type": "string"
     },
+    "license": {
+      "oneOf": [
+        {
+          "type": "string",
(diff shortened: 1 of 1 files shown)


olofk:
@kbroch-rivosinc I haven't checked how reuse and scancode-toolkit work in detail. It looks like they dig through all the source code checking for proper license info, but I'm not sure if or how we can make it check against the license tag in the core description file. The license tag isn't intended to cover the .core file itself, but the whole project.

FYI, our overall goal is working towards proper SPDX compliance for FuseSoC projects. This PR is a first step in that direction.

kbroch-rivosinc:
@olofk : Thanks for explanation.  Similar approach to this: https://packaging.python.org/en/latest/specifications/pyproject-toml/#license

PR #758, opened 2025-11-28
Add netrc support to allow users to work with private github repository cores

midisenpai:
This change adds support for private github FuseSoC core repositories.

To test out on a linux installation:
1. Create a .netrc file at `~/.netrc` with the format below. NOTE: The file permissions **MUST** be 600 for the .netrc file or it will not be used.
2. Attempt to run a FuseSoC core target for a core in a private repository.

Example `~/.netrc` file format:
```
machine github.com
    login your_github_username
    password ghp_your_classic_gh_token
```

A new token can be created in github at **Settings → Developer settings → Personal access tokens**. I've only tested with the classic github token type and the only scope necessary is the `repo` checkbox and it's children.

This change does not introduce any new dependencies for FuseSoC, does not introduce or modify any flags, and should be transparent (used when .netrc with 600 permissions is available and the github.com machine is present in the .netrc file, ignored otherwise).

### DIFF
--- a/fusesoc/provider/github.py
+++ b/fusesoc/provider/github.py
@@ -3,6 +3,7 @@
 # SPDX-License-Identifier: BSD-2-Clause
 
 import logging
+import netrc
 import os.path
 import sys
 import tarfile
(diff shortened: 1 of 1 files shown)


joshua-apex:
While the `.netrc` file is an option, you can also use `~/.ssh/config` and associate ssh keys with hostnames.

e.g. `~/.ssh/config` example:
```
Host github.com
  HostName github.com
  User git
  IdentityFile ~/.ssh/id_ed25519
```

With a config like this, all github.com based repositories accessible with your private key will be accessible within fusesoc.

Feel free to close this pull request, I believe this is an issue which does not require code changes to solve.

olofk:
Alright then. I'm closing. The less code, the better :)

PR #760, opened 2025-12-10
Expand the autotype filter to include commonly used file types

amykyta3:
This change expands the `autotype` filter to include several additional file extensions that are popular and unambiguous enough that they ought to be included.

Added the following file type filters:
* `*.vhd` or `*.vhdl` --> "vhdlSource" (I was surprised this was missing)
* `*.xci` --> "xci" (matches the existing file_type used in the Vivado flow)
* `*.bd` --> "bd" (matches the existing file_type used in the Vivado flow)
* `*.sdc` --> "SDC"
* `*.rdl` --> "systemRDL" (selfishly adding since I'll be building out some FuseSoC RDL tooling!)

I also organized the type map into groups to make it easier to maintain in the future.

### DIFF
--- a/fusesoc/filters/autotype.py
+++ b/fusesoc/filters/autotype.py
@@ -7,13 +7,32 @@ logger = logging.getLogger(__name__)
 class Autotype:
     def run(self, edam, work_root):
         type_map = {
+            # HDL sources
+            ".sv": "systemVerilogSource",
+            ".v": "verilogSource",
+            ".vhd": "vhdlSource",
+            ".vhdl": "vhdlSource",
+
+            # Vivado IP
+            ".xci": "xci",
+            ".bd": "bd",
+
+            # Verilator
+            ".vlt": "vlt",
+
+            # Design constraints
+            ".sdc": "SDC", # Synopsys-style
+            ".xdc": "xdc", # Vivado-style
+
+            # Software
             ".c": "cSource",
             ".cpp": "cppSource",
-            ".sv": "systemVerilogSource",
+
+            # Scripts
             ".tcl": "tclSource",
-            ".v": "verilogSource",
-            ".vlt": "vlt",
-            ".xdc": "xdc",
+
+            # Register map definitions
+            ".rdl": "systemRDL",
         }
         for f in edam["files"]:
             if not "file_type" in f:


rpatel3001:
hey @amykyta3 did any of that fusesoc + peakrdl tooling get published somewhere?

amykyta3:
Not yet. I'm still experimenting with what workflow makes most sense.

For `PeakRDL-regblock`, I will likely aim for a mechanism that adapts to FuseSoC's generators, since that seems like the best fit for peripheral CSR block generation. This may require some enhancements to FuseSoC's generator mechanism to make it more adaptable. Still working out some ideas.

I don't see too many good reasons to create an Edalize tool/flow for PeakRDL, aside from maybe an adapter that generates HTML. But at that point, I would probably prefer to call a FuseSoC API from Python to extract the list of RDL files so I can feed it into the SphinxDocs back-end with this plugin: https://sphinx-peakrdl.readthedocs.io/en/latest/index.html

rpatel3001:
cool, will be interested to see what becomes available. here's what I have hacked together in case any of it is useful. It's got some stuff specific to my environment so not immediately useful to everyone, but a decent starting point.

core file:
```
CAPI=2:
name: XX:common:peakrdl_v1_0_1
description: Generate FW and SW memory map source files from SystemRDL source with the PeakRDL toolchain

provider:
  name    : git
  repo    : https://XX.com/XX/peakrdl-XX.git
  version : v1.0.1

generators:
  gen_regmap:
    interpreter: python
    command: generate.py
    cache_type: none
    description: Generate FW and SW memory map source files from SystemRDL source with the PeakRDL toolchain
    usage: |
      Parameters:
        rdl_top (required): Path to top level RDL file
        logical_name (optional): Set to override the default logical_name, which is the `rdl_top` filename + _regmap
        html (optional): Set to any truthy value to generate HTML output. Can cause issues on Windows due to long path lengths.

targets:
  default:
    filesets:
      - rdl

filesets:
  rdl:
    file_type: user
    files:
      - field_types.rdl
```

generate.py:
```
#!/usr/bin/env python3

from fusesoc.capi2.generator import Generator
from os import chdir, getcwd
from pathlib import Path
from peakrdl.main import main as peakrdl
import sys

from unittest.mock import patch


class PeakRDLGenerator(Generator):
    def run(self):
        peakrdl_root = Path(__file__).parent
        rdl_top = self.config["rdl_top"]
        rdl_top_file = Path(self.files_root) / rdl_top
        default_lib = Path(rdl_top).stem + "_regmap"
        libname = self.config.get("logical_name", default_lib)
        vlnvarr = self.vlnv.split(":")
        vlnvarr[2] = libname
        self.vlnv = ":".join(vlnvarr)
        self.core_file = self.vlnv.split(":")[2] + ".core"
        self.description = f"Register map generated from {rdl_top} for library {libname}"

        exporters = ["vhdl", "excel"]
        if self.config.get("html"):
            exporters.append("html")

        gen_root = getcwd()
        chdir(peakrdl_root)
        for exporter in exporters:
            peakrdlargs = ["peakrdl", exporter, "field_types.rdl", rdl_top_file.as_posix(), "-o", (Path(gen_root)/f"build/{exporter}").as_posix()]
            with patch.object(sys, "argv", peakrdlargs):
                peakrdl()
        chdir(gen_root)

        files = []
        simfiles = []

        for f in Path("build/vhdl").iterdir():
            if "_tb" in f.stem:
                simfiles.append(f.as_posix())
            else:
                files.append(f.as_posix())

        self.add_files(files, file_type="vhdlSource-2008", logical_name=libname, targets=["default", "sim"])
        self.add_files(simfiles, file_type="vhdlSource-2008", logical_name=libname, fileset="tb", targets=["sim"])


if __name__ == "__main__":
    g = PeakRDLGenerator()
    g.run()
    g.write()
```

some notes:
* version is appended to name in VLNV, I do that for RTL cores to allow multiple versions to be used in one project without causing filename conflicts but it's probably not required for this core since only one version should exist per project.
* cache_type is none because the top level RDL file can have `` `include``s that wouldn't get analyzed by fusesoc for differences
* the `target` and `filesets` in the core are not really needed but work around this bug (https://github.com/olofk/edalize/issues/507)
* `description` is not included in the generated core file (https://github.com/olofk/fusesoc/issues/768)
* `vhdl` and `excel` are custom exporters and `field_types.rdl` contains some commonly used field types, these files are in the `provider` repo with generate.py

PR #762, opened 2026-01-06
fix(fusesoc.py): fixed missing 'tool' definition, catching correct error

Soarnsky:
Not sure if `ImportError` is the right exception to catch here, but `ToolResolutionError` seems more accurate


ArkoshEternal:
pre-commit failed. Please run pre-commit run --all-files to pass pipe lint

Issue #675, opened 2024-03-07
Feature request: ability to control arguments to Generators


shareefj:
I came across a use case today that I hadn't thought about before and is probably particular to our use case.  We're using a Generator to convert Migen modules to Verilog and then add them as dependencies.  I have a simple Python generator that takes, among other things, a list of parameters to use in the Migen instantiation.

So for example, imagine you have a configurable Migen module that takes `n_channel` as an instantiation argument.  This is passed to the generator via a core file that looks like:

```yaml
CAPI=2:
name: ::example

generate:
  verilog_netlist:
    generator: migen-verilog
    parameters:
      other:
        - stuff
        - here
      params:
        - name: n_channel
          type: int
          value: 2

filesets:
  generators:
    depend:
      - nq:utils:generators

targets:
  default: &default
    filesets:
      - generators
    generate:
      - verilog_netlist
```

What we'd like to do is enable parameterisation of these arguments from the command line and I started looking at how I could use FuseSoC parameters to do this.  I ended up modifying FuseSoC to pass the parameters defined in the core file down to the generator (by modifying arguments passed to Ttptttg and paths in between), and then realised after all of that that the overriding of parameters on the command line only get passed to the Edalize backend and aren't available inside of FuseSoC.

At least that was my understanding after a quick hack.

So is that correct and if so, can you think of any other way of overriding inputs to generators?

Addendum:
In my example, I ended up creating a new parameter type that was named `param` (not at all confusing) and inside my generator, I was passing in all core file parameters and extracting the matching one.  But it was getting the default value rather than the command line overridden one.

olofk:
I have thought about that too. Some generators could definitely benefit from being easily parameterized outside of the .core file. There also happens to be some other places that could benefit from the same mechanism as well.

One idea I have had for many years but never gotten around to implement is support for variables in core files. This is kind of an extension of use flags. The idea would be that you could have something like
```
generate:
  verilog_netlist:
    generator: migen-verilog
    parameters:
      other:
        - stuff
        - here
      params:
        - name: {{ n_channel }}
          type: int
          value: 2
```
and define that with, I don't know, `fusesoc run --var n_channel=2 ...` or something like that.

As a side note, did you know that you can specify generator parameters directly in the targets? You can check e.g. [here](https://github.com/olofk/corescore/blob/master/corescore.core#L584) for the syntax.

WisniewskiP:
[...]

PR #765, opened 2026-01-28
CAPI2: Variables

ArkoshEternal:
# CAPI2 Variables 
Introducing flag-like variables feature discussed originally in #675

## High Level Feature Description: 

Add the ability to assign CAPI2 entries to a variable, denoted by a string which starts with ``$``
Globally resolve variables either from defaults set in a target, or from CLI arguments 

## Change Details: 
* Add ``variables`` section to ``target`` which allows for the definition of int, string, or bool variables which can be defined as key-value pairs
* Add support for new meta-character, ``$``, which can be used in a manner similar to bash variables wherever a flag ternary operation can currently be used 
* Add new cli for ``fusesoc run``, ``--var key=value`` , which can supplement target held variables
* Add testing to ensure functionality of the variables infrastructure
* Add variables to build_system documentation 

## Design Decisions
* Add variables as coequal to flags instead of as redesigning flags as a data class to prevent any major re-writes of the flags system, although unifying is appealing at a later time
* Intentionally did not fuse them with parameters because of parameters isolation to use in the backend rather than the front end

## Pain Points Solved
* ** Allows for upstream control of downstream generators**
* Allows for less complex flag ternary trees downstream
* Simplifies duplicated tool options (e.g. same command passed to vlogan/vhdlan and vcs in vcs simulations)
* Gives end users more flexibility in implementing their vision

## Considered Extensions 
* Allow for inline variable defaults (e.g. my_key: $my_value:=<default_value>)

The implementation was intentionally simple, happy to iterate on it more if there's specific pain points that can be solved in this change.

### DIFF
--- a/doc/source/user/build_system/index.rst
+++ b/doc/source/user/build_system/index.rst
@@ -38,6 +38,7 @@ A full reference documentation on the CAPI2 core file format can be found in the
    filters.rst
    flags.rst
    generators.rst
+   variables.rst
    virtual_cores.rst
    mappings.rst
    hooks.rst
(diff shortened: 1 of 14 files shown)


ArkoshEternal:
Will be re approaching this by using the flags dict to manage the variables instead of a new class

Issue #755, opened 2025-11-06
`fusesoc tool list` command generates ImportError: cannot import name 'walk_tool_packages' from 'edalize.edatool'


kbroch-rivosinc:
```
kbroch@kbroch-mac.local:~ via 🐍 v3.13.6 (venv-test-fusesoc) took 5s
❯ fusesoc tool list
Traceback (most recent call last):
  File "/Users/kbroch/venv-test-fusesoc/bin/fusesoc", line 7, in <module>
    sys.exit(main())
             ~~~~^^
  File "/Users/kbroch/venv-test-fusesoc/lib/python3.13/site-packages/fusesoc/main.py", line 810, in main
    fusesoc(args)
    ~~~~~~~^^^^^^
  File "/Users/kbroch/venv-test-fusesoc/lib/python3.13/site-packages/fusesoc/main.py", line 800, in fusesoc
    args.func(fs, args)
    ~~~~~~~~~^^^^^^^^^^
  File "/Users/kbroch/venv-test-fusesoc/lib/python3.13/site-packages/fusesoc/main.py", line 226, in list_tools
    from edalize.edatool import get_edatool, walk_tool_packages
ImportError: cannot import name 'walk_tool_packages' from 'edalize.edatool' (/Users/kbroch/venv-test-fusesoc/lib/python3.13/site-packages/edalize/edatool.py)
```

reproducible by just creating a venv and installing fusesoc.  Here's versions I have:

```
kbroch@kbroch-mac.local:~ via 🐍 v3.13.6 (venv-test-fusesoc)
❯ python --version
Python 3.13.6

kbroch@kbroch-mac.local:~ via 🐍 v3.13.6 (venv-test-fusesoc)
❯ pip show edalize fusesoc
Name: edalize
Version: 0.6.3
Summary: Library for interfacing EDA tools such as simulators, linters or synthesis tools, using a common interface
Home-page: https://github.com/olofk/edalize
Author:
Author-email: Olof Kindgren <olof@award-winning.me>
License:
Location: /Users/kbroch/venv-test-fusesoc/lib/python3.13/site-packages
Requires: Jinja2
Required-by: fusesoc
---
Name: fusesoc
Version: 2.4.4
Summary: Award-winnning package manager and build abstraction tool for HDL code
Home-page: https://fusesoc.net
Author:
Author-email: Olof Kindgren <olof@award-winning.me>
License:
Location: /Users/kbroch/venv-test-fusesoc/lib/python3.13/site-packages
[...]

PR #769, opened 2026-03-20
Fix "fusesoc tool list" with edalize 0.6.2+

jmerdich:
Edalize removed the methods previously used to list the tools (`walk_tool_packages`). Use the new methods when present.

### DIFF
--- a/fusesoc/main.py
+++ b/fusesoc/main.py
@@ -216,19 +216,39 @@ def list_cores(fs, args):
 
 
 def list_tools(fs, args):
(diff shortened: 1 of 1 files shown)


jmerdich:
Fixes #755

olofk:
Thanks for this. Functionality looks fine, but the code doesn't pass the formatting check. Make sure you run `black` on the code to format this properly. Easiest way is to set it up as a pre-commit hook by running `pre-commit install` in your local repo. This will do a formatting check every time you commit.

PR #771, opened 2026-04-16
Corrected version examples under dependencies with requirement operat…

rschmidlin:
Small adaptation to documentation


olofk:
Whoops! Thanks Raul :)

PR #775, opened 2026-05-07
Fix #646: prefix hook script names with owning core's VLNV

### LINKED ISSUE #646: inherited scripts are overwritten
It seems that if you have a core file that creates a post_build hook with a script, lets say FOO. If you coincidentally create a post_build hook and script also called FOO, either by inheriting from another core file or creating it in the core file in question, this will create a makefile with several targets all with the same name, and hence only the latter target will be run.
since this is done in the make file, is it worth having fusesoc embed a part of the VLNV or some sort of identifier to ensure that all the hook targets are unique?

phverg:
## Problem
Fixes https://github.com/olofk/fusesoc/issues/646.

Edalize generates one Makefile target per hook script using the script's `name` field. When two cores in the dependency tree defined a hook script with the same user-facing name (typical when a child core inherits/overrides a hook from a parent), both ended up in the merged EDAM `hooks` list with identical names, so the Makefile got duplicate targets and only the last one was actually run — silently dropping the inherited hook. @olofk confirmed in the issue that the fix needs to be on the FuseSoC side.

## Solution
In `Edalizer.create_edam`, prefix each emitted hook script's `name` with the owning core's `sanitized_name`. `Core.get_scripts` keeps returning the user-facing name unchanged so other call sites and `test_capi2` expectations stay valid; the prefixing happens only at the EDAM emission boundary, where the owning-core context is available.

## Test plan
- [x] New two-core fixture in `tests/capi2_cores/hooks_collision/` and `test_hook_script_names_are_unique_per_core` in `tests/test_edalizer.py`. Verified the test fails without the fix and passes with it.
- [x] Full pytest suite green
- [x] `pre-commit run -a` clean

### DIFF
--- a/fusesoc/edalizer.py
+++ b/fusesoc/edalizer.py
@@ -253,8 +253,17 @@ class Edalizer:
             # Extract flow options
             snippet["flow_options"] = core.get_flow_options(_flags)
 
-            # Extract scripts
+            # Extract scripts. Prefix every hook script name with the owning
+            # core's sanitized VLNV: edalize generates one Makefile target per
+            # hook script using its `name`, so two cores defining a script
+            # called `foo` would otherwise produce duplicate targets and only
+            # the last one would run.
             snippet["hooks"] = core.get_scripts(rel_root, _flags)
+            for hook_scripts in snippet["hooks"].values():
+                for script in hook_scripts:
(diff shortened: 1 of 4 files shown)


olofk:
That makes sense. Thanks for fixing

PR #777, opened 2026-05-07
Fix #700: invalidate provider cache when its config changes

### LINKED ISSUE #700: Problem with provider version of a core and fuesoc cache
We found a problem regarding the field `version` in the section `provider` of a core and the fusesoc cache folder.

When the value in the field `version` is changed from `v1.0.4-rc0` to `v1.0.4-rc1` for example, fusesoc doesn't notice this and the files are not updated in the cache folder. The only workaround ist to change the version in the core `name: :some:core:1.0.4` => `name: :some:core:1.0.5` which is not possible, because the core is used as a dependency of another core and an update of the version number of the core would cause a domino effect.

We have a single repo for all our cores and the cleanest solution to this problem would be if fusesoc checks also the version in the provider section.

I think this issue should be classified as a bug.

```
CAPI=2:

name: :some:core:1.0.4
description: Some core

filesets:
    rtl:
        files:
            - src/some_file.sv

targets:
    default: &default
        filesets:
            - rtl

provider:
    name: git
    repo: <some_repo>
    version: v1.0.4-rc0

```

phverg:
## Problem
Fixes https://github.com/olofk/fusesoc/issues/700.

`Provider.status()` only checks whether the cache directory exists on disk. Once a core has been fetched, FuseSoC happily reuses the cached files for every subsequent run, even if the user has since changed the core's `provider:` block — most notably the `version` field. So flipping `version: v1.0.4-rc0` → `version: v1.0.4-rc1` silently keeps building against the rc0 sources, with the only workaround being to bump the core's own VLNV version (which then cascades through dependents).

## Solution
After a successful fetch, write a `.fusesoc-provider-config.json` marker inside the cache directory containing a normalised JSON copy of the provider config. `status()` reads it back on subsequent runs and flags the cache as `outofdate` whenever the current config no longer matches, which then triggers the existing clean-and-refetch path.

Caches written by older FuseSoC don't have the marker; treat those as valid for backward compatibility (no surprise refetch on upgrade) and let the next fetch write a fresh marker so any future drift is caught.

## Test plan
- [x] `test_provider_cache_invalidates_on_config_change` — confirms a changed `version` flips status from `downloaded` to `outofdate`.
- [x] `test_provider_legacy_cache_without_marker_is_valid` — confirms a pre-marker cache is still treated as `downloaded`.
- [x] Existing provider tests (`test_git_provider`, `test_url_provider`, `test_uncachable`, etc.) still pass.

### DIFF
--- a/fusesoc/provider/provider.py
+++ b/fusesoc/provider/provider.py
@@ -2,6 +2,7 @@
 # Licensed under the 2-Clause BSD License, see LICENSE for details.
 # SPDX-License-Identifier: BSD-2-Clause
 
+import json
 import logging
 import os
 import shutil
@@ -11,6 +12,12 @@ from importlib import import_module
 from fusesoc.utils import Launcher
 
 logger = logging.getLogger(__name__)
+
+# File written inside a cached files_root after a successful checkout. Holds a
+# normalised JSON copy of the provider's CAPI2 config so subsequent runs can
+# detect when the user changed e.g. the provider ``version`` and the cache
+# needs to be refreshed.
+_CONFIG_MARKER = ".fusesoc-provider-config.json"
 
 
 def get_provider(name):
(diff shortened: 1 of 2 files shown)


olofk:
This makes sense, but I need to think if there are some other implications of this. About to release a new FuseSoC version, so this fix will likely land after that.

PR #778, opened 2026-05-07
Fix #767: drop duplicates when merging '..._append' lists

### LINKED ISSUE #767: CAPI2 Syntax Gotcha: fileset_append existing fileset
# Background 
While converting to VCS 3-Step recently, I ran into a few file already defined error, despite the file only being defined once in the relevant fileset. I realized someone had extended a target and then appended a fileset that already existed in the base target. This was a simple syntax fix, but seems like it could be incorrect behavior. 

# Proposed Fix
I plan to implement either a parse error if this occurs, or a system which removes fileset duplication from within one core file. Will see which one fits better with the codebase, and if I automatically fix, add a warning that it occurred.

# Example "Problem" 

```yaml
CAPI=2: 

name: my:test:core 

filesets: 
  files_rtl: 
    - dut.sv
  file_type: systemVerilogSource

  files_dv: 
    - tb.sv
  file_type: systemVerilogSource 

targets:
  default: &default
    filesets: [files_rtl]
  
  sim_base: &sim_base: 
    <<: *default
    filesets_append: [files_dv]
    flow: sim
    flow_options: 
      tool: vcs
    toplevel: tb

  unit_test: 
    <<: *sim_base
   # issue occurs on next line
   filesets_append: [files_dv]
   parameters: 
     TESTNAME=unit_test

parameters: 
   TESTNAME: 
     datatype: string
     paramtype: plusarg

```

phverg:
## Problem
Fixes https://github.com/olofk/fusesoc/issues/767.

A natural inheritance pattern using YAML merge plus `..._append`:

```yaml
sim_base: &sim_base
  filesets_append: [files_dv]
  ...
unit_test:
  <<: *sim_base
  filesets_append: [files_dv]
```

…leaves `unit_test` with `filesets_append: [files_dv, files_dv]` after YAML's list merge. Every file in `files_dv` then ends up listed twice in the emitted EDAM, and the EDA backend rejects the build with a `file already defined` error. The same happens for `files_append` re-listing a file already present in `files`. The natural inheritance pattern is effectively unusable.

## Solution
Deduplicate when extending the base list inside `CoreData._append_lists`. Items that are already present (in the base list or earlier in the same `..._append`) are dropped; new items are still appended in order. This keeps the existing append-only and base-only test cases unchanged.

## Test plan
- [x] Extended `tests/capi2_cores/misc/append.core` with three new targets covering the duplicate cases (in-base, within-append, append-only).
- [x] Extended `test_capi2_append` to assert duplicates are dropped and order preserved; existing assertions still pass.
- [x] Manually verified the issue scenario (YAML merge + double `filesets_append`): EDAM file list now contains each file once.
- [x] Full `pytest` suite green locally.

### DIFF
--- a/fusesoc/capi2/core.py
+++ b/fusesoc/capi2/core.py
@@ -40,7 +40,7 @@ class CoreInterface:
     def __post_init__(self):
         parsed_capi = self.parser.read(self.core_file)
 
-        self.handle = CoreHandle.from_dict(parsed_capi)
+        self.handle = CoreHandle.from_dict(parsed_capi, self.core_file)
 
         self.name = Vlnv(self.get_data({}).name)
 
--- a/fusesoc/capi2/core_handle.py
+++ b/fusesoc/capi2/core_handle.py
@@ -73,8 +73,11 @@ class CoreHandle:
     _cache: dict[FlagDefs, Core[str]] = field(init=False, default_factory=dict)
 
     @classmethod
-    def from_dict(cls, capi_dict: dict) -> "CoreHandle":
-        return cls(Core[Expr].model_validate(capi_dict))
+    def from_dict(cls, capi_dict: dict, core_file=None) -> "CoreHandle":
+        # core_file is passed as validation context so schema validators can
+        # point warnings at the core file they originate from.
+        context = {"core_file": str(core_file)} if core_file else None
+        return cls(Core[Expr].model_validate(capi_dict, context=context))
 
     def get(self, flag_defs: FlagDefs) -> Core[str]:
         if core := self._cache.get(flag_defs):
--- a/fusesoc/capi2/schema/common.py
+++ b/fusesoc/capi2/schema/common.py
@@ -1,13 +1,16 @@
 # SPDX-License-Identifier: BSD-2-Clause
 # SPDX-FileCopyrightText: FuseSoC contributors
 
+import logging
 from collections.abc import Mapping
 from typing import Any, Generic, Literal, TypeVar
 
-from pydantic import BaseModel, ConfigDict, Field, model_validator
+from pydantic import BaseModel, ConfigDict, Field, ValidationInfo, model_validator
 from typing_extensions import Self
 
 from ..exprs import Expr
+
+logger = logging.getLogger(__name__)
 
 
 class FrozenModel(BaseModel):
@@ -35,15 +38,30 @@ class FrozenModel(BaseModel):
 ExprOrStr = TypeVar("ExprOrStr", str, Expr)
 
 
-def _merge_append_keys(data: Any, keys: tuple[str, ...]) -> Any:
-    """Merge `<key>_append` lists into `<key>` lists at parse time."""
+def _merge_append_keys(
+    data: Any, keys: tuple[str, ...], info: ValidationInfo | None = None
+) -> Any:
+    """Merge `<key>_append` lists into `<key>` lists at parse time.
+
+    Items already present in the base list (or repeated within the
+    `<key>_append` list itself) are dropped with a warning, so a fileset
+    or file pulled in twice through target inheritance is only emitted
+    once (see #767)."""
     if not isinstance(data, dict):
         return data
+    core_file = (info.context or {}).get("core_file") if info else None
     for key in keys:
         ak = f"{key}_append"
         if ak in data:
             base = list(data.get(key, []))
-            base.extend(data.pop(ak))
+            for item in data.pop(ak):
+                if item in base:
+                    source = f" in {core_file}" if core_file else ""
+                    logger.warning(
+                        f"Dropping duplicate entry '{item}' from '{ak}'{source}"
+                    )
+                else:
+                    base.append(item)
             data[key] = base
     return data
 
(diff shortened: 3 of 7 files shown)


olofk:
I understand what this fixes, but thinking about it I can't see how on earth we would get into this situation. The provided example kind of undermines the whole idea of inheritance as the point is that you would _not_ have to add the fileset in the target that inherits. I realize your example is intentionally contrived, but could you (or someone else) come up with a more realistic situation where this could happen?

ArkoshEternal:
> I understand what this fixes, but thinking about it I can't see how on earth we would get into this situation. The provided example kind of undermines the whole idea of inheritance as the point is that you would _not_ have to add the fileset in the target that inherits. I realize your example is intentionally contrived, but could you (or someone else) come up with a more realistic situation where this could happen?

I opened this issue in the first place. What I ran into was a situation in which the VCS two-step backend was masking the overinclusion of filesets (in this case, some like dv_base_pkg.sv file). An engineer with more familiarity with yaml inheritance I think would have caught this, but people who don't want to take the time to understand it and more or less copy-paste targets, can get themselves into a confusing situation with this. 

I think this is more or less defensive programming against stupid mistakes that don't get highlighted by every EDA backend. I spent a good few hours looking at the filelists edalize generated trying to figure out how two of the same toplevel files were getting included in the flists before realizing it was an issue with accidental dual inclusion of filesets in a core file.

fusesoc/capi2/coredata.py
```diff
@@ -55,6 +55,18 @@ def _expand_use(self, data, flags):
...
+    def _extend_unique(base, extra):
+        """Extend ``base`` with each item from ``extra`` that isn't already
+        present, preserving order. Dropping duplicates here keeps the same
+        fileset/file from being emitted twice when YAML-merged target
+        inheritance combines two ``..._append`` lists, which would otherwise
+        trip EDA backends' duplicate-file checks (see #767)."""
+        for item in extra:
+            if item not in base:
+                base.append(item)
+        return base
```
ArkoshEternal: In the original issue I opened, I think we resolved to throw a warning when this occurs since it is not good CAPI2 syntax. 

If you can instance a logger and throw a warning here (or wherever makes the most sense) which points to the source of the warning so its easily fixable, I think that more or less closes the original issue I opened (if Olof agrees)

fusesoc/capi2/coredata.py
```diff
@@ -55,6 +55,18 @@ def _expand_use(self, data, flags):
...
+
     def _append_lists(self, data):
```
ArkoshEternal: Thinking more on this, might be outside the scope of this PR, but I feel like it might make sense to use a set() item instead of lists to ensure we don't have duplication within an individual core, but I think we might be able to address that in some typing refactor

ArkoshEternal (changes requested):
Request to add warning throws. See logging behavior in a file like main.py if you need inspiration

fusesoc/capi2/coredata.py
ArkoshEternal: This is really a note for myself, will resolve it here

tymonx:
This is related with this issue https://github.com/yaml/pyyaml/issues/165. Most YAML parsers should keep only the last defined key (override). It is a common approach to override existing keys when using YAML anchors (for example, GitLab CI).

Possible alternative solution https://github.com/yaml/pyyaml/issues/165#issuecomment-3689109906 to resolve that when parsing YAML file.

phverg:
Thanks for the review! Added the warning — and in the meantime the branch needed a rebase: main's type-checked core schema refactor (96516b4) removed `CoreData`, where both the original fix and the first version of the warning lived, so the change is now reimplemented on top of the new pydantic schema:

- Duplicates are dropped in `_merge_append_keys` (`fusesoc/capi2/schema/common.py`), with a warning per dropped duplicate, e.g. `Dropping duplicate entry 'fs2' from 'filesets_append' in /path/to/append.core` — it names the item, the offending `..._append` key, and the core file, so it's directly actionable. The core file reaches the schema validators via pydantic validation context (passed in `CoreHandle.from_dict`).
- `test_capi2_append` asserts the warnings are emitted (one per dropped duplicate) and checks the message text.
- I verified the duplicate issue still reproduces on current main before reapplying the fix.

On the `set()` idea — agreed that's probably better left for a typing refactor; I kept lists here to preserve append order and keep the diff scoped.

@tymonx I looked into the pyyaml angle. From what I can tell, FuseSoC may not go through pyyaml's `<<` handling here — `Inheritance.yaml_merge_2_fusesoc_merge` appears to rewrite the merge operator before parsing, and FuseSoC then does the merge itself. If I'm reading `merge_dict(concat_list_appends_only=True)` right, child keys already override parent keys much like you describe, with `..._append` lists being the one exception — they're concatenated, I believe intentionally, so a child target can extend an inherited list. A parse-time approach could probably work too, though I think it would need to hook into FuseSoC's own merge step rather than the YAML loader. Two things I wasn't sure it would cover in this particular bug: the repeated `filesets_append` keys end up in different mappings (parent anchor vs. child), and the case from the original report — a fileset already in the inherited `filesets`, re-listed in `filesets_append` — doesn't involve a key collision at all. That's why I went with dedup-with-warning at the merge step, where both cases pass through one place.

PR #780, opened 2026-05-11
add network switch to tests

HU90m:
This allows the user to filter out tests requiring network access.

```sh
pytest -m "not network"
```


ArkoshEternal (approved):
I think a further extension of this would be a variable in the CI that can run the pipes in airgapped networks in the case people need to develop on internal forks before being able to commit upstream.

PR #781, opened 2026-05-11
Add static and runtime type checking to the CAPI2 Schema.

HU90m:
I've made these changes such that the interface of `fusesoc.core.Core` remains unchanged. This was done in the interest of keeping changes minimal.

We can change this interface, and make use of the CAPI2 objects, as we expand type checking into the rest of the repository in later PRs.

As discussed, I've dumped the python version. Python 3.9 is end-of-life and 3.10 offers many ergonomic features. See: https://github.com/olofk/fusesoc/pull/734

Pydantic has been used for runtime type checking, which saves a lot of schema validation logic. Pydantic can be used to generate the jsonschema. A future PR could use `sphinx-autodoc-pydantic` to generate nice documentation, remove use of the old jsonschema and provide the jsonschema generated from Pydantic for people creating tools in other languages. This may be of interest to you @amykyta3

Both mypy and ty have been added to pre-commit for static type checking.

Beyond the normal test suite, I also checked for changes in the output of:

```bash
fusesoc --cores-root=/tmp/repos/serv run \
  --setup --build-root=/tmp/build \
  --target=verilator_tb \
  servant
fusesoc --cores-root=/tmp/repos/opentitan/hw run \
  --setup --build-root=/tmp/build \
  --target=lint --flag=tool_ascentlint \
  --flag=files_rtl_tapeout \
  --mapping=lowrisc:systems:top_earlgrey:0.1 \
  --mapping=lowrisc:prim_generic:all:0.1 \
  lowrisc:systems:chip_earlgrey_asic
```

*NB: https://github.com/olofk/fusesoc/commit/e7c243b98777840fa010524d4c7cd464843463ab broke some opentitan core files, so has to be reverted whenever testing against opentitan. Something to be aware of when updating FuseSoC in opentitan @machshev*


ArkoshEternal (approved):
Since this is majorly non-breaking, and checking would require that we have active fusesoc runs to check against edge-cases that are not in the existing test framework, it may make sense to add the local checks into the CI for fusesoc. 

I think we should merge this in as an initial set and then we may have to make revisions after we do more integration into the rest of the codebase

PR #783, opened 2026-05-17
Fix #751: include the full generator definition in the cache hash

### LINKED ISSUE #751: Generator caching logic
In the caching logic of generators I think that the command being executed should be accounted for the generation of the hash fingerprint, along with its [YAML input](https://github.com/olofk/fusesoc/blob/d9b2ed38040f3cb0da79c8730783efd8e4c67e83/fusesoc/edalizer.py#L559) and [input files](https://github.com/olofk/fusesoc/blob/d9b2ed38040f3cb0da79c8730783efd8e4c67e83/fusesoc/edalizer.py#L565).

Am I missing some cases where this is not desirable?

phverg:
## Problem
Fixes https://github.com/olofk/fusesoc/issues/751.

`Ttptttg._sha256_input_yaml_hexdigest` hashes only `{parameters, vlnv, gapi}` from `generator_input`. Editing the generator — whether by pointing `command:`/`interpreter:` at a different file or by editing the script's bytes in place — leaves the hash identical, so subsequent runs silently serve the previously-cached output even though the program that produced it has changed.

## Solution
Fold the generator definition into the hash: `command`, `interpreter`, and a SHA-256 of the resolved script bytes. A clean rebuild now happens whenever any of these changes.

Why slightly broader than the literal ask in the issue: "the command being executed" naturally includes the script content, not just the declared path. Hashing only the declared `command:` string catches renames but not in-place script edits — which is the common "I updated the generator and the cache still served stale output" symptom. Hashing the resolved bytes closes that gap and mirrors the pattern PR #777 uses for provider patch bytes.

**Deliberately NOT hashed:**
- The resolved interpreter path from `shutil.which(interpreter)`. That depends on the system PATH at run time and is not part of the generator definition the user signs in their `.core` file. The interpreter name is what we sign.
- The generator's `root` path. It's an absolute path that differs between machines and checkouts, which would make cache directory names environment-specific and the test reference hashes machine-specific. Two generators with the same relative `command` rooted under different directories are still distinguished by the script bytes hash: if their bytes differ, the cache differs; if their bytes match, sharing the cache is the desired behaviour.

This necessarily invalidates existing on-disk caches for cacheable generators on first run after upgrade (the directory name embeds the hash). Users get one extra regeneration and consistent behaviour afterwards. The hardcoded reference hashes in `test_generators` are updated accordingly.

## Test plan
- [x] `test_generator_input_hash_includes_command` — proves changing `command` flips the hash; stable across repeats.
- [x] `test_generator_input_hash_includes_interpreter` — proves swapping `python3` → `python2` flips the hash.
- [x] `test_generator_input_hash_includes_script_bytes` — missing → v1 → v2 → v1-again sequence, proving the hash depends on bytes not on mtime.
- [x] Existing `test_generators` reference hashes updated; full `tests/test_edalizer.py` green (7 passed).

### DIFF
--- a/fusesoc/edalizer.py
+++ b/fusesoc/edalizer.py
@@ -572,31 +572,19 @@ class Ttptttg:
         }
 
     def _sha256_input_yaml_hexdigest(self):
+        """Hash the generator inputs plus the generator definition itself.
+
+        The command, interpreter, and a hash of the script bytes are folded
+        in so editing the generator invalidates the cache. The resolved
+        interpreter path and the generator ``root`` are excluded to keep the
+        hash portable across machines and checkouts.
+        """
         data = self.generator_input.copy()
         # Remove files_root since that is not deterministic
         data.pop("files_root")
-        # Fold the generator definition into the hash so that updating the
-        # generator -- its declared command/interpreter *or* the script's
-        # bytes -- actually invalidates the cache. Without this a changed
-        # generator with unchanged user-facing inputs silently serves
-        # stale output. See #751.
-        #
-        # Deliberately NOT hashed:
-        #
-        # * The resolved interpreter path from ``shutil.which(interpreter)``
-        #   -- depends on the system PATH at run time and is not part of
-        #   the generator definition the user signs in their .core file.
-        # * The generator's ``root`` path -- it's an absolute path that
-        #   differs between machines/checkouts, which would make hashes
-        #   non-portable and cache directory names environment-specific.
-        #   Two generators with the same relative ``command`` rooted under
-        #   different directories are still distinguished here by
-        #   ``__generator_script_hash__``: if their script bytes differ,
-        #   the hash differs; if their bytes match, sharing the cache is
-        #   the desired behaviour.
-        data["__generator_command__"] = self.generator.get("command", "")
-        data["__generator_interpreter__"] = self.generator.get("interpreter", "")
-        data["__generator_script_hash__"] = self._generator_script_sha256()
+        data["generator_command"] = self.generator.get("command", "")
+        data["generator_interpreter"] = self.generator.get("interpreter", "")
+        data["generator_script_hash"] = self._generator_script_sha256()
         return hashlib.sha256(utils.yaml_dump(data).encode()).hexdigest()
 
     def _generator_script_sha256(self):
--- a/fusesoc/edalizer.py
+++ b/fusesoc/edalizer.py
@@ -575,7 +575,50 @@ class Ttptttg:
         data = self.generator_input.copy()
         # Remove files_root since that is not deterministic
         data.pop("files_root")
+        # Fold the generator definition into the hash so that updating the
+        # generator -- its declared command/interpreter *or* the script's
+        # bytes -- actually invalidates the cache. Without this a changed
+        # generator with unchanged user-facing inputs silently serves
+        # stale output. See #751.
+        #
+        # Deliberately NOT hashed:
+        #
+        # * The resolved interpreter path from ``shutil.which(interpreter)``
+        #   -- depends on the system PATH at run time and is not part of
+        #   the generator definition the user signs in their .core file.
+        # * The generator's ``root`` path -- it's an absolute path that
+        #   differs between machines/checkouts, which would make hashes
+        #   non-portable and cache directory names environment-specific.
+        #   Two generators with the same relative ``command`` rooted under
+        #   different directories are still distinguished here by
+        #   ``__generator_script_hash__``: if their script bytes differ,
+        #   the hash differs; if their bytes match, sharing the cache is
+        #   the desired behaviour.
+        data["__generator_command__"] = self.generator.get("command", "")
+        data["__generator_interpreter__"] = self.generator.get("interpreter", "")
+        data["__generator_script_hash__"] = self._generator_script_sha256()
         return hashlib.sha256(utils.yaml_dump(data).encode()).hexdigest()
+
+    def _generator_script_sha256(self):
+        """SHA-256 of the resolved generator script bytes, or ``None`` if
+        the script cannot be read.
+
+        The ``None`` sentinel is intentional: a script that is unreadable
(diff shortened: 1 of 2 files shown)


fusesoc/edalizer.py
```diff
@@ -575,8 +575,51 @@ def _sha256_input_yaml_hexdigest(self):
         data = self.generator_input.copy()
         # Remove files_root since that is not deterministic
         data.pop("files_root")
+        # Fold the generator definition into the hash so that updating the
```
ArkoshEternal: Stylistic request: 

Please reduce the length of this comment to match with the rest of the file, or consider adding a docstring comment to the function which more concisely describes what is being hashed and why

fusesoc/edalizer.py
```diff
@@ -575,8 +575,51 @@ def _sha256_input_yaml_hexdigest(self):
...
+        #   the generator definition the user signs in their .core file.
+        # * The generator's ``root`` path -- it's an absolute path that
+        #   differs between machines/checkouts, which would make hashes
+        #   non-portable and cache directory names environment-specific.
+        #   Two generators with the same relative ``command`` rooted under
+        #   different directories are still distinguished here by
+        #   ``__generator_script_hash__``: if their script bytes differ,
+        #   the hash differs; if their bytes match, sharing the cache is
+        #   the desired behaviour.
+        data["__generator_command__"] = self.generator.get("command", "")
```
ArkoshEternal:
```suggestion
        data["generator_command"] = self.generator.get("command", "")
```

fusesoc/edalizer.py
```diff
@@ -575,8 +575,51 @@ def _sha256_input_yaml_hexdigest(self):
...
+        data["__generator_interpreter__"] = self.generator.get("interpreter", "")
```
ArkoshEternal:
```suggestion
        data["generator_interpreter"] = self.generator.get("interpreter", "")
```

fusesoc/edalizer.py
```diff
@@ -575,8 +575,51 @@ def _sha256_input_yaml_hexdigest(self):
...
+        data["__generator_script_hash__"] = self._generator_script_sha256()
```
ArkoshEternal:
```suggestion
        data["generator_script_hash"] = self._generator_script_sha256()
```

fusesoc/edalizer.py
```diff
@@ -575,8 +575,51 @@ def _sha256_input_yaml_hexdigest(self):
...
         return hashlib.sha256(utils.yaml_dump(data).encode()).hexdigest()
 
+    def _generator_script_sha256(self):
+        """SHA-256 of the resolved generator script bytes, or ``None`` if
```
ArkoshEternal, line 591: Please clean up comment to be more concise

tests/test_edalizer.py
```diff
@@ -338,3 +338,146 @@ def test_hook_script_names_are_unique_per_core():
...
+            }
+        }
+        t = Ttptttg(ttptttg, fake_core, generators, str(tmp_path / "work"))
+        return t._sha256_input_yaml_hexdigest()
+
+    assert hash_with_interp("python3") != hash_with_interp("python2")
+
+
+def test_generator_input_hash_includes_script_bytes(tmp_path):
+    """Editing the generator script in place -- same path, same
```
ArkoshEternal, line 425: Reduce comment verbosity to describe what is being tested and not why the the thing needs to be tested given other changes in the PR

tests/test_edalizer.py
```diff
@@ -338,3 +338,146 @@ def test_hook_script_names_are_unique_per_core():
     assert len(set(names)) == 2, f"hook script names collided: {names}"
     assert "hookcollision-parent_0_myhook" in names
     assert "hookcollision-child_0_myhook" in names
+
+
+def test_generator_input_hash_includes_command(tmp_path):
```
ArkoshEternal: Comment language: Describe feature which is being tested instead of a spec mandate. language like "must" is a bit out of place.

tests/test_edalizer.py
```diff
@@ -338,3 +338,146 @@ def test_hook_script_names_are_unique_per_core():
...
+    h_new = hash_with_cmd("gen_v2.py")
+    h_repeat = hash_with_cmd("gen_v1.py")
+
+    assert h_old != h_new, "changing generator command must invalidate the cache hash"
+    assert h_old == h_repeat, "same command must produce stable hash"
+
+
+def test_generator_input_hash_includes_interpreter(tmp_path):
+    """Swapping the interpreter (e.g. ``python3`` → ``python2``) is also a
+    real environment change that should invalidate the cache."""
```
ArkoshEternal:
```suggestion
    invalidates the cache."""
```

ArkoshEternal (changes requested):
Code mostly looks fine - completely agree with the notion as this also makes working on a generator a lot easier. I ran into this issue recently and am glad to see a fix up for it. 

General notion on testing: tests should name what they are testing, not the properties that should exist/etc.

phverg:
Thanks for the review! Addressed all the comments in the latest push (24f1edc):

- Renamed the hash keys to `generator_command` / `generator_interpreter` / `generator_script_hash` (dropped the dunders).
- Condensed the long block comment into a function docstring describing what's hashed and what's deliberately excluded.
- Reworded the test names to describe the behaviour under test rather than the property (`test_changing_generator_command_invalidates_cache_hash`, etc.), and dropped the "must"/spec-mandate language from the docstrings and assertion messages.
- Recomputed the reference hashes in `test_generators` for the renamed keys.

`pytest tests/test_edalizer.py` is green and `pre-commit run -a` is clean.

ArkoshEternal:
Thank you for so promptly addressing the review. It looks good to me, just going to request an optional review & then merge

@olofk This looks good to me, if you want to give it a once over feel free else I'm comfortable with merging. I have run into this issue and believe it is a good fix.

ArkoshEternal (approved):

PR #787, opened 2026-06-17
Increase scope of ty

erikwa:
- Increase scope of ty (from fusesoc/capi2/ to fusesoc/)
- Fix a couple of issues found by ty
- Add ty overrides to pyproject.toml
- Optional: Remove mypy in favor of ty

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -41,23 +41,6 @@ repos:
       - id: ruff-check
       # The formatter could be run in the future
       #- id: ruff-format
-  - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v1.13.0
-    hooks:
-      - id: mypy
-        # Typed runtime deps mypy needs to resolve imports for the
-        # strict-checked modules. Keep in sync with `pyproject.toml`'s
-        # `[project].dependencies` for any package that ships type info.
-        additional_dependencies:
-          - pydantic>=2.13.3
-          - typing_extensions
-          - pyparsing>=2.3.1
-          - types-PyYAML
-        args: [--config-file=pyproject.toml]
-        # Don't let pre-commit pass file paths as arguments — they would
-        # override `files = [...]` in pyproject.toml and pull every Python
-        # file into the check regardless of the strict whitelist.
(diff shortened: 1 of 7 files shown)


erikwa:
Argh... Guess I missed some overrides that are needed. Will fix.

ArkoshEternal:
From the ty readme: 

ty uses 0.0.x versioning. ty does not yet have a stable API; breaking changes, including changes to diagnostics, may occur between any two versions. See the https://github.com/astral-sh/ty/issues/1889 tracking issue for a detailed overview of currently supported features.

Very big fan of how fast this is vs mypy, but I am slightly concerned over expanding its use given its lack of support.

@HU90m I probably should have brought this up on the initial PR where you introduced ty, but figure I'll bring it up here now since we're increasing its scope. 

Actual python changes look fine tho 👍

ArkoshEternal (commented):
see above comment, shoulda left it in review

might also make sense to just use both for the time being if they don't conflict

HU90m:
Really nice to see the type checking scope expand. Yes, ty is experimental. We should almost definitely have mypy or pyright. They're both super feature complete and stable. We're a small project so the speed is a non-issue for CI.

I had added ty because I was using the LSP. Pyrefly also looks great. Both have fast and incremental LSPs which is really nice to have. I guess the question is whether we want multiple type checkers, i.e. also new type checkers in CI.

erikwa:
Thanks for the feedback. I agree that it is probably too early to rely solely on ty as our type checker.

I initially looked into this because my Neovim LSP setup uses ty and ruff, and it was picking up the reduced scope from the pyproject.toml file. While I could probably work around this in my local Neovim configuration, I figured it would be cleaner to have the tools use the project's configured scope by default.

If we still want to keep the pre-commit hook scoped strictly to fusesoc/capi2/, we can achieve that by adding an args parameter to the hook configuration:

_args: ["fusesoc/capi2/"]_


I did some additional work today to get mypy passing on fusesoc/, and it's almost complete. Let's close this PR, and I'll submit a new one that handles both mypy and ty. We can decide later whether to keep the ty pre-commit hook.

PR #788, opened 2026-06-22
Type checker fixes

erikwa:
- Increase project scope of mypy and ty to fusesoc/
- Update mypy and ty versions used by pre-commit
- Fix issues detected by mypy and ty
- Limit pre-commit scope of ty to fusesoc/capi2/

Maybe we should remove the pre-commit hook for ty completely? Lets discuss.

### DIFF
--- a/fusesoc/capi2/core.py
+++ b/fusesoc/capi2/core.py
@@ -200,7 +200,7 @@ class CoreInterface:
 
     def get_scripts(self, files_root, flags):
         self._debug(f"Getting hooks for flags '{str(flags)}'")
-        hooks = {}
+        hooks: dict[str, list[dict[str, Any]]] = {}
         for hook, scripts in self._get_script_names(flags).items():
             hooks[hook] = []
             for script in scripts:
@@ -420,7 +420,7 @@ class CoreInterface:
     def get_ttptttg(self, flags):
         self._debug(f"Getting ttptttg for flags {str(flags)}")
         target_name, target = self._get_target(flags)
-        ttptttg = []
+        ttptttg: list[dict[str, Any]] = []
 
         if not target:
             return ttptttg
@@ -456,7 +456,7 @@ class CoreInterface:
         return ttptttg
 
     def _get_vpi(self, flags):
-        vpi = {}
+        vpi: dict[str, Any] = {}
         target_name, target = self._get_target(flags)
         if not target:
             return vpi
--- a/fusesoc/capi2/core.py
+++ b/fusesoc/capi2/core.py
@@ -132,7 +132,7 @@ class CoreInterface:
                         shutil.copytree(src, dst, dirs_exist_ok=True)
 
         # Clean out leftover files from previous builds
-        for root, dirs, files in os.walk(dst_dir):  # ty: ignore[invalid-assignment]
+        for root, _, files in os.walk(dst_dir):
             for f in files:
                 _abs_f = os.path.join(root, f)
                 _rel_f = os.path.normpath(os.path.relpath(_abs_f, dst_dir))
--- a/fusesoc/edalizer.py
+++ b/fusesoc/edalizer.py
@@ -10,7 +10,7 @@ import pathlib
 import shutil
 from filecmp import cmp
 from importlib import import_module
-from typing import Sequence
+from typing import Any, Sequence
 
 from fusesoc import utils
 from fusesoc.capi2.coreparser import Core2Parser
@@ -62,6 +62,7 @@ class Edalizer:
         self.resolve_env_vars = resolve_env_vars
 
         self.generators = {}
+        self.edam: dict[str, Any] = {}
 
         self._resolved_or_generated_cores = []
 
@@ -208,12 +209,12 @@ class Edalizer:
                 )
 
     def create_edam(self):
-        first_snippets = []
-        snippets = []
-        last_snippets = []
-        parameters = {}
+        first_snippets: list[dict[str, Any]] = []
+        snippets: list[dict[str, Any]] = []
+        last_snippets: list[dict[str, Any]] = []
+        parameters: dict[str, Any] = {}
         for core in self.cores:
-            snippet = {}
+            snippet: dict[str, Any] = {}
 
             logger.debug(f"Collecting EDAM parameters from {str(core.name)}")
             _flags = self._core_flags(core)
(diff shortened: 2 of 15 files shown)


fusesoc/capi2/core.py
```diff
@@ -132,7 +132,7 @@ def export(self, dst_dir, flags={}):
                         shutil.copytree(src, dst, dirs_exist_ok=True)
 
         # Clean out leftover files from previous builds
-        for root, dirs, files in os.walk(dst_dir):  # ty: ignore[invalid-assignment]
+        for root, _subdirs, files in os.walk(dst_dir):
```
olofk: If we're not using `_subdirs`, we can change this to `_` instead.

fusesoc/filters/custom.py
```diff
@@ -8,7 +8,7 @@
     from yaml import CDumper as Dumper
     from yaml import CLoader as Loader
 except ImportError:
-    from yaml import Dumper, Loader
+    from yaml import Dumper, Loader  # type: ignore[assignment]
```
olofk, line 11: Out of curiosity...what's the actual issue here?

fusesoc/filters/spdxgen.py
```diff
@@ -150,18 +156,19 @@ def _generate_package(self, work_root, ci_id, vlnv, license_expr, flist):
                     "@id": hash_id,
                 },
             ]
+            assert isinstance(file_rel["to"], list)  # Make ty happy
```
olofk: Can we solve this differently? If this might be triggered, I would prefer to have an exception instead that we can catch and inform the user. If it's never supposed to be triggered then it feels like a false positive and it would be nicer to fix the root cause.

fusesoc/utils.py
```diff
@@ -9,13 +9,15 @@
 import warnings
 
 import yaml
+import yaml.parser
+import yaml.scanner
```
olofk, line 13: Why do we need to import these explicitly? Is it for Parser/ScannerError?

fusesoc/utils.py
```diff
@@ -9,13 +9,15 @@
...
 
 try:
     from yaml import CSafeDumper as YamlDumper
     from yaml import CSafeLoader as YamlLoader
 except ImportError:
-    from yaml import SafeDumper as YamlDumper
-    from yaml import SafeLoader as YamlLoader
+    from yaml import SafeDumper as YamlDumper  # type: ignore[assignment]
+    from yaml import SafeLoader as YamlLoader  # type: ignore[assignment]
```
olofk, line 20: Can anyone inform me why this needs a waiver?

olofk (changes requested):
Looks good aside from some minor things. Also have some questions. Happily surprised that there weren't more changes needed.

fusesoc/capi2/core.py
erikwa: True... fixed.

fusesoc/filters/custom.py
erikwa: From mypys static perspective "Loader" is first assigned CLoader type and "Dumper" is assigned CDumper type. Then "Loader" is assigned Loader type and "Dumper" is assigned Dumper type. The try and except are in the same scope and the Loader/CLoader classes and Dumper/CDumper classes are two different stand-alone classes that do not share any inherited relationship even though the API they expose might be equivalent.

fusesoc/utils.py
erikwa: See comment above about Loader/CLoader and Dumper/CDumper.

fusesoc/utils.py
erikwa: I did not give it that much thought since ty recommended it:

```shell
warning[possibly-missing-submodule]: Submodule `parser` might not have been imported                      --> fusesoc/utils.py:174:13
    |
174 |     except (yaml.parser.ParserError, yaml.scanner.ScannerError) as e:
    |             ^^^^^^^^^^^
    |
help: Consider explicitly importing `yaml.parser`

warning[possibly-missing-submodule]: Submodule `scanner` might not have been imported
   --> fusesoc/utils.py:174:38
    |
174 |     except (yaml.parser.ParserError, yaml.scanner.ScannerError) as e:
    |                                      ^^^^^^^^^^^^
    |
help: Consider explicitly importing `yaml.scanner`
``` 

I guess it is not able to find evidence that guarantees that yaml.parser and yaml.scanner exists in this namespace.

One thing I noticed though when I looked at this again was that when running ty through pre-commit it was never detected. After a little bit more digging I noticed that the difference was that the pre-commit environment for ty did not have types-PyYAML as dependency, which my local enivoronment had. I will add it to the pre-commit deps section for ty for later even if we decide to comment out the pre-commit ty check.

fusesoc/filters/spdxgen.py
erikwa: Changed the implementation to solve it without the assert. Check the latest push :)

fusesoc/utils.py
ArkoshEternal: mypy also has issues with the types-PyYAML but it explicitly warns it as an error. 

Do we want to run pre-commit with --install-types for the static linters?

fusesoc/edalizer.py
```diff
@@ -62,6 +62,7 @@ def __init__(
         self.resolve_env_vars = resolve_env_vars
 
         self.generators = {}
+        self.edam: dict[str, Any] = {}
```
ArkoshEternal, line 65: Just making a note here that eventually this should be a pydantic data class ala the capi which we can integrate whenever I get to that :)

pyproject.toml
```diff
@@ -80,10 +80,13 @@ indent-width = 4
...
 
 [tool.mypy]
+python_version = "3.10"
 mypy_path = "."
 namespace_packages = true
 explicit_package_bases = true
-files = ["fusesoc/capi2/"]
+check_untyped_defs = true
+enable_error_code = ["deprecated"]
+files = ["fusesoc/"]
```
ArkoshEternal, line 91:
```suggestion
plugins = [
    "pydantic.mypy"
]
```

ArkoshEternal: suggesting we use the pydantic mypy plugins to enable pydantic checking instead of of waivering pydantic

ArkoshEternal (changes requested):

ArkoshEternal:
> Looks good aside from some minor things. Also have some questions. Happily surprised that there weren't more changes needed.

There are going to be a lot more if we're up to enforce no-Any typing eventually, but I think this is a really really good start

pyproject.toml
erikwa: Thanks, I was not aware of the pydantic plugin for mypy. Will add it.

fusesoc/utils.py
erikwa: From what I can see mypy has that option, but ty check does not. Did we come to a conclusion if we should keep the ty pre-commit check or not?

olofk: Let's keep it like this for now

olofk (approved):
Excellent!

PR #791, opened 2026-06-30
feat: add list type when passing options from command line to Edalize flow backend

### LINKED ISSUE #790: feat: add a new list type when passing options from command line to Edalize flow backend
Context: https://github.com/olofk/edalize/pull/537#issuecomment-4842115269

https://github.com/olofk/fusesoc/blob/f15e1c8a76815c4f391231dd0e743e2b683c6b45/fusesoc/edalizer.py#L334-L341

Edalize flow backends are expecting list of arguments (aka list of strings) and core files are using YAML array with YAML strings. But FuseSoC is passing that to Edalize as a string and Edalize flow backend is defining a tool option like `xrun_options` as a ...string.

We need a new type `list` (or `args` or `array`) that will allow convert passed options from command line to flow backend as a list of strings. And use `str` for strings only.

tymonx:
Closes https://github.com/olofk/fusesoc/issues/790

Context: https://github.com/olofk/edalize/pull/537#issuecomment-4842115269

This PR is needed to pass tool arguments directly from `fusesoc` command line interface to Edalize backend flow. Example:

```plaintext
fusesoc run --target sim <core> --xrun_options="-input script.tcl -access +rwc"
```

This was not supported because `fusesoc` wasn't able to handle a list of arguments. Supported types were limited to `bool`, `str`, `int`, `real` or `file`.

This PR:

* Added a new type `list` to handle arguments (including dashes, quotes and whitespace characters) that will be passed to Edalize backend
* Removed unnecessary `nargs=1`. They should not be used when passing a single value because it is hard to distinguish a non-list value from a list. Additionally, this also removes unnecessary brackets `[]` around metavars when invoking the help
* Added `metavar` to inform users about expected value type. By default, the Python ArgumentParser is using upper-case name of option. It only duplicates a name without providing any value in return for end users
* Replaced duplicated `dictype` with the global `_DICTTYPE` 

Added an unit test to test all variants:

```plaintext
pytest -k 'test_edalizer_parse_args'
```

All files were linted with the pre-commit.

### DIFF
--- a/fusesoc/edalizer.py
+++ b/fusesoc/edalizer.py
@@ -7,6 +7,7 @@ import hashlib
 import logging
 import os
 import pathlib
+import shlex
 import shutil
 from filecmp import cmp
 from importlib import import_module
@@ -23,10 +24,10 @@ logger = logging.getLogger(__name__)
 
 class FileAction(argparse.Action):
     def __call__(self, parser, namespace, values, option_string=None):
-        path = os.path.expandvars(values[0])
+        path = os.path.expandvars(values)
         path = os.path.expanduser(path)
         path = os.path.abspath(path)
-        setattr(namespace, self.dest, [path])
+        setattr(namespace, self.dest, path)
 
 
 def str2bool(v):
@@ -38,6 +39,22 @@ def str2bool(v):
         return False
     else:
         raise argparse.ArgumentTypeError("Boolean value expected.")
+
+
+#: Conversion table between Edalize types and FuseSoC command line arguments
+#: The ``metavar`` is used to inform users about expected value type
+_TYPEDICT_PARAMETERS = {
+    "bool": {"type": str2bool, "nargs": "?", "const": True, "metavar": "yes|no"},
+    "file": {"type": str, "action": FileAction, "metavar": "FILE"},
+    "int": {"type": int, "metavar": "INTEGER"},
+    "str": {"type": str, "metavar": "ARG"},
+    "real": {"type": float, "metavar": "NUMBER"},
+}
+
+_TYPEDICT_OPTIONS = {
+    **_TYPEDICT_PARAMETERS,
+    "list": {"type": shlex.split, "metavar": "ARGS"},
+}
 
 
 class Edalizer:
@@ -332,13 +349,6 @@ class Edalizer:
             merge_dict(self.edam, snippet)
 
     def _build_parser(self, backend_class, edam):
-        typedict = {
-            "bool": {"type": str2bool, "nargs": "?", "const": True},
-            "file": {"type": str, "nargs": 1, "action": FileAction},
-            "int": {"type": int, "nargs": 1},
-            "str": {"type": str, "nargs": 1},
-            "real": {"type": float, "nargs": 1},
-        }
         progname = "fusesoc run {}".format(edam["name"])
 
         parser = argparse.ArgumentParser(prog=progname, conflict_handler="resolve")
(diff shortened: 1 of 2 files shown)


olofk:
Unless I have misunderstood something, this should already be supported. Using e.g. [serv](https://github.com/olofk/serv/) as an example I can run `fusesoc run --target=verilator_tb servant --verilator_options="-threads 2 -j 4"` and in the EDAM I will get the following argument list
```
...
  verilator_options:
  - --trace
  - -threads
  - '2'
  - -j
  - '4'
...
```
where `--trace` is defined in the .core file and the rest of the args are supplied on the command line and chopped up into a list because `verilator_options` has `"list": True` defined in the edalize backend. My assumption is that we only need to set the same for xrun_options in the xcelium backend (which should had been set already).

Or am I missing something here?

tymonx:
@olofk I have changed the type in the Edalize `edalize/tools/xcelium.py` from the `str` to the `list`. Traceback when invoking:

```plaintext
fusesoc run --target <flow-sim-xcelium> <core> --xrun_options="-verbose"
```

Output:

```plaintext
  File ".venv/bin/fusesoc", line 10, in <module>
    sys.exit(main())
             ^^^^^^
  File ".venv/lib/python3.12/site-packages/fusesoc/main.py", line 836, in main
    fusesoc(args)
  File ".venv/lib/python3.12/site-packages/fusesoc/main.py", line 826, in fusesoc
    args.func(fs, args)
  File ".venv/lib/python3.12/site-packages/fusesoc/main.py", line 388, in run
    edam_file, backend = fs.get_backend(core, flags, args.backendargs)
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".venv/lib/python3.12/site-packages/fusesoc/fusesoc.py", line 172, in get_backend
    edalizer.parse_args(backend_class, backendargs)
  File ".venv/lib/python3.12/site-packages/fusesoc/edalizer.py", line 522, in parse_args
    parser = self._build_parser(backend_class, self.edam)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".venv/lib/python3.12/site-packages/fusesoc/edalizer.py", line 410, in _build_parser
    **typedict[v["type"]],
      ~~~~~~~~^^^^^^^^^^^
KeyError: 'list'
```

Issue:

https://github.com/olofk/fusesoc/blob/f15e1c8a76815c4f391231dd0e743e2b683c6b45/fusesoc/edalizer.py#L411

Because the `typedict` dictionary doesn't contain the `list` entry.

I'm using the Xcelium flow:

```yaml
targets:
  sim:
    flow: sim
    flow_options:
      tool: xcelium
```

> Using e.g. [serv](https://github.com/olofk/serv/) as an example I can run

Serv is not using the legacy tool mode :thinking: 

I'm also using the latest FuseSoC + Edalize from the `main` branches.

tymonx:
Oh my god. There is a separate `"list": True` switch between FuseSoC and Edalize...

https://github.com/olofk/edalize/blob/7e5f16f7a3c29b3ae744e1f19fa6ea67530edd70/edalize/tools/verilator.py#L41

```python
        "verilator_options": {
            "type": "str",
            "desc": "Additional options for verilator",
            "list": True,
        },
```

Today I learned something new. I will try this with `xcelium`.

tymonx:
@olofk Sorry for that. I have added the `"list": True` to the xcelium backend and it worked. I was misguided with this example:

https://github.com/olofk/edalize/blob/7e5f16f7a3c29b3ae744e1f19fa6ea67530edd70/edalize/flows/f4pga.py#L39

```python
            "vpr_options": {
                "type": "list",
                "desc": "Options to VPR, if the standard F4PGA values are not sufficient",
            },
```

And also it looked more natural to have `"type": "list"`. We can close this PR if you want. There are some valid changes that could be merged anyway like removing `nargs=1` and more unit tests.

I will add the `"list": True` setting to the https://github.com/olofk/edalize/pull/537.

PR #798, opened 2026-07-30
Support git submodules in libraries

wallento:


olofk:
My personal opinion is that the submodules should be added as separate libraries instead if possible. I'm a bit worried about how to handle all the potential error cases e.g. when doing `fusesoc library update`. But I also understand that some projects are set up to expect things to be located in locations relative to each other so let's pull this in, but it's kind of use at your own risk.

PR #799, opened 2026-07-31
Support default "tool" and "flow" flags when using flow API

mxsparks:
Fixes an issue where the default `tool_TOOLNAME` flags were never being set if not using the tool API. Also adds a default `flow_FLOWNAME` flag.

Passes ruff and tox (no new unit tests added).

### DIFF
--- a/fusesoc/main.py
+++ b/fusesoc/main.py
@@ -373,6 +373,23 @@ def run(fs, args):
         logger.error(str(e))
         exit(1)
 
+    try:
+        flow = core.get_flow(flags)
+        if flow and "flow" not in flags:
+            flags["flow"] = flow
+
+            flow_options = core.get_flow_options(
+                {"is_toplevel": True, **flags}
+            )
+            if flow_options and "tool" in flow_options and "tool" not in flags:
+                flags["tool"] = flow_options["tool"]
+    except SyntaxError as e:
+        logger.error(str(e))
+        exit(1)
+    except RuntimeError as e:
+        logger.error(str(e))
+        exit(1)
+
     # Unconditionally clean out the work root on fresh builds
     # if we use the old tool API or clean flag is set
     if do_configure and (not core.get_flow(flags) or args.clean):
--- a/doc/source/user/build_system/flags.rst
+++ b/doc/source/user/build_system/flags.rst
@@ -33,6 +33,9 @@ When running ``fusesoc run`` some flags are automatically made available.
 * ``target_TARGETNAME``: A flag that is set if a particular target is being built.
   ``TARGETNAME`` is replaced with the name of the target that is being built.
   For example, if the target ``synth`` is being built, the flag ``target_synth`` will be set.
+* ``flow_FLOWNAME``: A flag that is set when a particular flow is being run.
+  ``FLOWNAME`` is replaced with the name of the flow that is being run.
+  For example, if the simulation flow is being run, the flag ``flow_sim`` will be set.
 
 User-defined flags
 ~~~~~~~~~~~~~~~~~~


fusesoc/main.py
```diff
@@ -373,6 +373,23 @@ def run(fs, args):
...
+                {"is_toplevel": True, **flags}
+            )
+            if flow_options and "tool" in flow_options and "tool" not in flags:
+                flags["tool"] = flow_options["tool"]
+    except SyntaxError as e:
+        logger.error(str(e))
+        exit(1)
+    except RuntimeError as e:
+        logger.error(str(e))
+        exit(1)
```
tymonx, line 391: This could be simplify:

```suggestion
    except (SyntaxError, RuntimeError) as e:
        logger.error(str(e))
        exit(1)
```

fusesoc/main.py
```diff
@@ -373,6 +373,23 @@ def run(fs, args):
...
+    try:
+        flow = core.get_flow(flags)
+        if flow and "flow" not in flags:
+            flags["flow"] = flow
+
+            flow_options = core.get_flow_options(
+                {"is_toplevel": True, **flags}
+            )
+            if flow_options and "tool" in flow_options and "tool" not in flags:
+                flags["tool"] = flow_options["tool"]
```
tymonx, line 385: How about setting the `tool` flag based on the `flow` if the `tool` is not present 🤔 For example, `vivado` flow (because it is using the Vivado tool):

```suggestion
            if "tool" not in flags:
                flags["tool"] = flow_options["tool"] if flow_options and "tool" in flow_options else flow
```

olofk:
This is actually by design. The tool_TOOLNAME flags don't make sense for the flow API as there can be flows using multiple tools. The intention is to deprecate the `--tool` switch completely once we have migrated away from the tool API. It's a bit inconvenient when using a mix of tool and flow backends, but can be mitigated by passing `--flag=tool_TOOLNAME` on the command-line and/or setting `flags: {tool: TOOLNAME}` in the targets.

Adding an implicit flow_FLOWNAME flag could still be considered, but I would argue that it is better to set an explicit flag in the relevant targets to convey the intention more clearly. e.g. instead of checking that the flow is sim, the actual thing we want to do is to use a simulation model or decrease some reset timer. Perhaps we also want to use the shortened reset timer in a formal flow, or do some simulations without the simulation model. That also helps with reuse, when the core files are perhaps used in a new flow that the original author didn't consider.

tymonx:
>  The intention is to deprecate the --tool switch completely

@olofk But there is also the `--tool` switch after the core name that is part of the Flow API. If you look into proposed changes:

```python
if flow_options and "tool" in flow_options and "tool" not in flags:
    flags["tool"] = flow_options["tool"]
```

It is using the `--tool` argument from the Flow API not from the Tool API. There is something like `sim` flow that can have different tools like `xcelium`, `xsim` or `ghdl`. Having something that will automatically set the tool flag is handy for the Flow API as well because for example you could include/exclude Xilinx `unisim` libraries:

```yaml
files:
  unisim:
    depend:
      - xilinx:unisim:vcomponents

targets:
  default:
    filesets:
      - '!tool_vivado ? (!tool_xsim ? (unisim))'
    flow: sim
    flow_options:
      tool: xcelium
```

In case of Xilinx `unisim` libraries, you want to include them in all HDL simulators except the Xilinx XSim simulator or Vivado implementation tool.

Then user can select simulator in the `sim` Flow API with correct filesets:

```plaintext
fusesoc run --target sim <vlnv> --tool=<xsim|xcelium|ghdl|...>
```

I have similar changes like @mxsparks is proposing in my forked version of FuseSoC. Otherwise handling simulation libraries when using the `sim` Flow API will be nightmare. Xilinx/AMD Vivado/Xsim doesn't like providing `unisim` or `xpm` libraries explicitly but this is required when using other simulators.

mxsparks:
> Otherwise handling simulation libraries when using the sim Flow API will be nightmare.

Agreed. We currently have a mix of tool and flow API usage, and the current behavior requires me to revisit every flow-API target in several *hundred* dependent cores and add a redundant `flags: {tool: TOOLNAME}` key when the `flow_options: {tool: TOOLNAME}` key is already present.

PR #802, opened 2026-08-09
feat(librarymanager): allow to use custom providers

### LINKED ISSUE #797: bug(library): cannot use custom provider for library management
This is because of this `if` statement:

https://github.com/olofk/fusesoc/blob/d7490780e7435abc00b9f9a63351c7ea530c64e5/fusesoc/librarymanager.py#L23-L28

In my opinion it is unnecessary because later there is:

https://github.com/olofk/fusesoc/blob/d7490780e7435abc00b9f9a63351c7ea530c64e5/fusesoc/librarymanager.py#L55

This should already thrown an error.

**Use case:** we want to add a custom provider for git monorepo by using the [git sparse-checkout ](https://git-scm.com/docs/git-sparse-checkout) command to checkout specific version of FuseSoC core in the same monorepo.

tymonx:
Closes https://github.com/olofk/fusesoc/issues/797

This PR allows to use custom providers to manage FuseSoC libraries.

Additionally:

* Moved the `Library` class to own `fusesoc/library.py` module. This is required to avoid circular imports between provider and library manager. The implementation of the `Library.update()` method was moved to the library manager to break the circular import of the `Library` class
* Added missing type annotations in providers, library and library manager
* Added missing docstrings in providers, library and library manager
* The library manager is now abstracted from providers (except for the `local` provider). Replaced strings like `checkout` that are git specific with `update/initialize`
* Removed the `library_root` attribute from the library manager. It was never used anywhere. Dead attribute
* Optimized the `LibraryManager.get_library()` method for fast lookup by using dictionary instead of list
* Using lazy evaluation in logging. Reason: https://news.ycombinator.com/item?id=44600537

### DIFF
--- a/fusesoc/coremanager.py
+++ b/fusesoc/coremanager.py
@@ -416,11 +416,7 @@ class CoreManager:
     def __init__(self, config, library_manager=None):
         self.config = config
         self.db = CoreDB()
-        self._lm = (
-            LibraryManager(config.library_root)
-            if library_manager is None
-            else library_manager
-        )
+        self._lm = LibraryManager() if library_manager is None else library_manager
         self.core2parser = Core2Parser(
             config.resolve_env_vars_early, config.allow_additional_properties
         )
--- a/fusesoc/fusesoc.py
+++ b/fusesoc/fusesoc.py
@@ -25,7 +25,7 @@ class Fusesoc:
     def __init__(self, config):
         self.config = config
 
-        self.lm = LibraryManager(config.library_root)
+        self.lm = LibraryManager()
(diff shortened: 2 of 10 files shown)


olofk:
Thanks! This looks like a nice cleanup. The original code was a bit messy so I haven't looked too closely at the logic changes. We should probably have some more tests here to help future changes, but that can be done separately.

A note about library_root. The intention was to provide a way to set the default library location (i.e. `fusesoc_libraries`) from the command-line or configuration file. It looks like this was never wired up properly though, so I think it's the right call to drop it completely.

PR #806, opened 2026-08-13
docs: add homebrew installation in installation.rst

hongjr03:
I have contributed a new formula for FuseSoC to Homebrew (see [Homebrew/homebrew-core#298565](https://github.com/Homebrew/homebrew-core/pull/298565)) today. As Homebrew is the most widely used package manager for macOS users, I think adding a Homebrew installation section to the documentation would be helpful.


doc/source/user/installation.rst
```diff
@@ -126,3 +126,9 @@ You can use ``pip`` to install as same as Linux.
 .. code-block:: shell-session
 
    $ pip3 install --upgrade --user fusesoc
+
+You can also use ``homebrew`` to install FuseSoC.
```
ArkoshEternal: pref something like (feel free to customize):

`A FuseSoC formula is also available for ``homebrew`` users:` 

as a guard against people (windows users) who don't know what brew is

ArkoshEternal (commented):
@olofk 

Might be worth adding a CI step to keep the brew version up to date(?) 

Haven't published to brew much so just kinda going on vibes

doc/source/user/installation.rst
hongjr03: Okay, I've fixed the sentence.

hongjr03:
> Might be worth adding a CI step to keep the brew version up to date(?)

I think we don't need extra CI for this. Homebrew already autobumps formulae in homebrew-core via BrewTestBot, see https://docs.brew.sh/Autobump. For example, 2.4.7 was bumped automatically in https://github.com/Homebrew/homebrew-core/pull/303164.

ArkoshEternal (approved):

PR #807, opened 2026-08-18
Skip cores with errors in the dependencies during solve

erikwa:
Fix CoreDB.solve() not skipping cores when get_depends() raises SyntaxError. Previously, execution continued after logging the skip message, causing an exception for the first core or causing subsequent cores to inherit the previous cores dependencies.


olofk:
Oops! Good catch

PR #808, opened 2026-08-19
Pass flags to get_virtual() when evaluating the conflict map

erikwa:
Pass flags to get_virtual() when evaluating the conflict map

### DIFF
--- a/fusesoc/coremanager.py
+++ b/fusesoc/coremanager.py
@@ -247,7 +247,7 @@ class CoreDB:
     def solve(self, top_core, flags):
         return self._solve(top_core, flags)
 
-    def _get_conflict_map(self):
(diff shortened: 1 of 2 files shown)


olofk:
I haven't considered the case where a virtual core is the toplevel. Unless someone is asking specifically for it, I would say we disallow it to avoid the risk of opening a can of worms. Passing flags makes sense though.

erikwa:
> I haven't considered the case where a virtual core is the toplevel. Unless someone is asking specifically for it, I would say we disallow it to avoid the risk of opening a can of worms. Passing flags makes sense though.

Thanks, I will drop the is_toplevel commit for now and we can revisit this later if needed :)

PR #812, opened 2026-09-12
add devcontainer

ArkoshEternal:
add a development container & corresponding vscode instructions for fusesoc development. Based off of python3.12 devcontainer, also installs pre-commit automatically using a post-run script to save time on first commit to make it as seamless as possible.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,72 @@
+# fusesoc dev container
+FROM mcr.microsoft.com/devcontainers/python:3.12
+
+# extraneous package manifest:
+#   subversion
+#   openssh-client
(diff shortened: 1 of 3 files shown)


ArkoshEternal:
@erikwa can you try this out and see if it works for you?

erikwa:
> @erikwa can you try this out and see if it works for you?

Sure. Will do.

erikwa:
> > @erikwa can you try this out and see if it works for you?
> 
> Sure. Will do.

Tried the devcontainer and it seems to work as expected. It will be a nice addition. 

The only thing I think is missing (except for a small clarification on a comment) is a section about this in the devsetup document. I created a PR on your fork with a draft suggestion.

